PR #4, opened 2022-11-04
Add function parameters

tonowak:


src/rustdoc_schema.graphql
```diff
@@ -462,6 +471,9 @@ type Function implements Item & FunctionLike & Importable {
   unsafe: Boolean!
   async: Boolean!
 
+  # edges from FunctionLike
+  parameters: [FunctionParameter!]
```
obi1kenobi: A couple minor nitpicks:
- Let's try to keep the schema elements in order parallel to the implemented interfaces followed by the "own" elements (properties / edges). This keeps it a bit more consistent and easier to read: `Item`'s elements always before `FunctionLike`'s ones, own at the end, etc.
- Let's keep edge names singular, because when used in a query, they only capture more than one element if we use an aggregation like `@fold`.

```suggestion
  # edges from FunctionLike
  parameter: [FunctionParameter!]
```

src/adapter.rs
```diff
@@ -961,7 +1019,7 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                 None => Box::new(std::iter::empty()),
                                 Some(token) => {
                                     let origin = token.origin;
-                                    let enum_item = token.as_enum().expect("token was not a Enum");
+                                    let enum_item = token.as_enum().expect("token was not an Enum");
```
obi1kenobi, line 1018: 🎉

src/adapter.rs
```diff
@@ -855,6 +887,32 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                     (ctx, neighbors)
                 }))
             }
+            "Function" | "Method" if matches!(edge_name.as_ref(), "parameters") => {
+                let _current_crate = self.current_crate;
+                let _previous_crate = self.previous_crate;
```
obi1kenobi:
```suggestion
```
We don't seem to be using these, so let's not keep them around.

src/adapter.rs
```diff
@@ -855,6 +887,32 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
...
+                    let neighbors: Box<dyn Iterator<Item = Self::DataToken> + 'a> =
+                        match &ctx.current_token {
+                            None => Box::new(std::iter::empty()),
+                            Some(token) => {
+                                let origin = token.origin;
+                                let decl = if current_type_name.as_ref() == "Function" {
+                                    &token.as_function().expect("token was not a Function").decl
+                                } else {
+                                    &token.as_method().expect("token was not a Method").decl
+                                };
```
obi1kenobi: As another option:
```suggestion
                                let decl = token.as_function().map(|f| &f.decl).unwrap_or(&token.as_method().expect("token was neither a Function nor a Method").decl);
```

The key difference is that the current code assumes that `current_type_name.as_ref() != "Function"` implies `current_type_name.as_ref() == "Method"`. This is currently true, because there are no edges that point directly to `FunctionLike`, but it's also a bit of a trap and a bug waiting to happen as soon as that sort of edge is added to the schema.

Recall that `current_type_name` here is the statically-known type of the vertex based on the query which includes `FunctionLike` as an option, not the specific type of a specific vertex which must be either `Function` or `Method` and can't be `FunctionLike`.

src/adapter.rs
```diff
@@ -855,6 +887,32 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                     (ctx, neighbors)
                 }))
             }
+            "Function" | "Method" if matches!(edge_name.as_ref(), "parameters") => {
```
obi1kenobi:
```suggestion
            "Function" | "Method" | "FunctionLike" if matches!(edge_name.as_ref(), "parameter") => {
```
Changes:
- Make `parameter` singular -- more detail in another comment below.
- Add `FunctionLike` since `current_type_name` is the statically-determined type bound based on the query, just to prevent a future bug -- also more detail in another comment below.

obi1kenobi (commented):
Looks good 🎉 Just a couple of small nitpicks that should be quick to apply.

src/rustdoc_schema.graphql
```diff
@@ -441,6 +441,15 @@ interface FunctionLike {
   const: Boolean!
   unsafe: Boolean!
   async: Boolean!
+  parameter: [FunctionParameter!]
```
obi1kenobi, line 446:
```suggestion

  # own edges
  parameter: [FunctionParameter!]
```

obi1kenobi (approved):
Awesome 🚀

PR #5, opened 2022-11-05
New schema for attributes

mgr0dzicki:
Introducing new schema for attributes in order to make the handling of tricky cases (like [obi1kenobi/cargo-semver-check#104](https://github.com/obi1kenobi/cargo-semver-check/issues/104)) in queries easier.

This is a draft of changes for rustdoc v21, support for other versions will be added once this is completed.

The code is already working, but some issues need to be discussed:
- [x] The instances of `Attribute` and `AttributeValue` are owned by tokens and therefore cloned each time a new token is created. How can we improve this? Move the ownership somewhere else? Use `Rc`?
- [x] TryFrom impl for AttributeValue panics if the given string is not in one of three recognized forms. They seem to cover all possible cases, but are we sure about that? Maybe it would be better to somehow ignore it (for example returning AttributeValue with only as_string set), so that the adapter will not fail on some attributes which are indifferent to us?
- [x] Ignore white spaces or not (especially in `as_string`)? For example, when the attribute is `#[repr(C, u8)]`, the second argument will have `as_string=' u8'`, but `base='u8'`. 
- [x] Naming. I like the names of properties and edges, but I am not really sure about `AttributeValue`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trustfall-rustdoc-adapter"
-version = "21.1.0"
+version = "21.2.0"
 edition = "2021"
 authors = ["Predrag Gruevski <obi1kenobi82@gmail.com>"]
 license = "Apache-2.0 OR MIT"
@@ -11,7 +11,7 @@ readme = "./README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-trustfall_core = "0.0.4"
+trustfall_core = "0.0.7"
 rustdoc-types = "0.17.0"
 anyhow = "1.0.58"
 regex = "1.6.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,9 +13,6 @@ readme = "./README.md"
 [dependencies]
 trustfall_core = "0.0.7"
 rustdoc-types = "0.17.0"
-anyhow = "1.0.58"
-regex = "1.6.0"
-lazy_static = "1.4.0"
 
 [dev-dependencies]
 serde_json = "1.0.85"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,9 @@ readme = "./README.md"
 [dependencies]
 trustfall_core = "0.0.4"
 rustdoc-types = "0.17.0"
+anyhow = "1.0.66"
+regex = "1.6.0"
+lazy_static = "1.4.0"
 
 [dev-dependencies]
 serde_json = "1.0.85"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ readme = "./README.md"
 [dependencies]
 trustfall_core = "0.0.4"
 rustdoc-types = "0.17.0"
-anyhow = "1.0.66"
+anyhow = "1.0.58"
 regex = "1.6.0"
 lazy_static = "1.4.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trustfall-rustdoc-adapter"
-version = "21.2.0"
+version = "21.3.0"
 edition = "2021"
 authors = ["Predrag Gruevski <obi1kenobi82@gmail.com>"]
 license = "Apache-2.0 OR MIT"
@@ -11,7 +11,7 @@ readme = "./README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-trustfall_core = "0.0.7"
+trustfall_core = "0.1.1"
 rustdoc-types = "0.17.0"
 
 [dev-dependencies]
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -102,6 +102,13 @@ impl Origin {
             kind: TokenKind::ImplementedTrait(path, trait_def),
         }
     }
+
+    fn make_function_parameter_token<'a>(&self, name: &'a str) -> Token<'a> {
+        Token {
+            origin: *self,
+            kind: TokenKind::FunctionParameter(name),
+        }
+    }
 }
 
 #[non_exhaustive]
@@ -124,6 +131,7 @@ pub enum TokenKind<'a> {
     Attribute(Attribute),
     AttributeValue(AttributeValue),
     ImplementedTrait(&'a Path, &'a Item),
+    FunctionParameter(&'a str),
 }
 
 #[allow(dead_code)]
@@ -166,6 +174,7 @@ impl<'a> Token<'a> {
                 rustdoc_types::Type::Primitive(..) => "PrimitiveType",
                 _ => "OtherType",
             },
+            TokenKind::FunctionParameter(..) => "FunctionParameter",
         }
     }
 
@@ -255,6 +264,13 @@ impl<'a> Token<'a> {
             rustdoc_types::ItemEnum::Function(func) => Some(func),
             _ => None,
         })
+    }
+
+    fn as_function_parameter(&self) -> Option<&'a str> {
+        match &self.kind {
+            TokenKind::FunctionParameter(name) => Some(name),
+            _ => None,
+        }
     }
 
     fn as_method(&self) -> Option<&'a Method> {
@@ -436,6 +452,17 @@ fn get_function_like_property(token: &Token, field_name: &str) -> FieldValue {
     }
 }
 
+fn get_function_parameter_property(token: &Token, field_name: &str) -> FieldValue {
+    let function_parameter_token = token
+        .as_function_parameter()
+        .expect("token was not a FunctionParameter");
+
+    match field_name {
+        "name" => function_parameter_token.into(),
+        _ => unreachable!("FunctionParameter property {field_name}"),
+    }
+}
+
 fn get_impl_property(token: &Token, field_name: &str) -> FieldValue {
     let impl_token = token.as_impl().expect("token was not an Impl");
     match field_name {
@@ -595,6 +622,9 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                         property_mapper(ctx, field_name.as_ref(), get_function_like_property)
                     }))
                 }
+                "FunctionParameter" => Box::new(data_contexts.map(move |ctx| {
+                    property_mapper(ctx, field_name.as_ref(), get_function_parameter_property)
+                })),
                 "Impl" => {
                     Box::new(data_contexts.map(move |ctx| {
                         property_mapper(ctx, field_name.as_ref(), get_impl_property)
@@ -888,6 +918,30 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                     (ctx, neighbors)
                 }))
             }
+            "Function" | "Method" | "FunctionLike" if matches!(edge_name.as_ref(), "parameter") => {
+                Box::new(data_contexts.map(move |ctx| {
+                    let neighbors: Box<dyn Iterator<Item = Self::DataToken> + 'a> = match &ctx
+                        .current_token
+                    {
+                        None => Box::new(std::iter::empty()),
+                        Some(token) => {
+                            let origin = token.origin;
+                            let decl = token.as_function().map(|f| &f.decl).unwrap_or_else(|| {
+                                &token
+                                    .as_method()
+                                    .expect("token was neither a Function nor a Method")
+                                    .decl
+                            });
+
+                            Box::new(decl.inputs.iter().map(move |(name, _type_)| {
+                                origin.make_function_parameter_token(name)
+                            }))
+                        }
+                    };
+
+                    (ctx, neighbors)
+                }))
+            }
             "Struct" => match edge_name.as_ref() {
                 "field" => {
                     let current_crate = self.current_crate;
@@ -994,7 +1048,7 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                 None => Box::new(std::iter::empty()),
                                 Some(token) => {
                                     let origin = token.origin;
-                                    let enum_item = token.as_enum().expect("token was not a Enum");
+                                    let enum_item = token.as_enum().expect("token was not an Enum");
 
                                     let item_index = match origin {
                                         Origin::CurrentCrate => &current_crate.inner.index,
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1317,9 +1317,7 @@ mod tests {
 
     #[test]
     fn rustdoc_json_format_version() {
-        let version = rustdoc_types::FORMAT_VERSION;
-        let current_crate =
-            load_rustdoc_from_file(Path::new(&format!("./test_data/rustdoc_v{version}.json")));
+        let current_crate = load_rustdoc_from_file(Path::new("./test_data/rustdoc_v21.json"));
 
         assert_eq!(current_crate.format_version, rustdoc_types::FORMAT_VERSION);
     }
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -129,7 +129,7 @@ pub enum TokenKind<'a> {
     ImplementedTrait(&'a Path, &'a Item),
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Attribute {
     is_inner: bool,
     content: AttributeValue,
@@ -159,7 +159,7 @@ impl<'a> TryFrom<&'a String> for Attribute {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct AttributeValue {
     as_string: String,
     base: String,
@@ -194,7 +194,6 @@ impl<'a> TryFrom<&'a String> for AttributeValue {
             let mut i = 0;
             let mut depth = 0;
             let ref arg_str = captures["arguments"];
-            println!("{}", arg_str);
             let mut arguments: Vec<AttributeValue> = Vec::new();
             let mut only_white = true;
             for (j, c) in arg_str.chars().enumerate() {
@@ -1372,6 +1371,8 @@ mod tests {
     use trustfall_core::{frontend::parse, interpreter::execution::interpret_ir, ir::FieldValue};
 
     use crate::indexed_crate::IndexedCrate;
+    use crate::adapter::Attribute;
+    use crate::adapter::AttributeValue;
 
     use super::RustdocAdapter;
 
@@ -1416,55 +1417,82 @@ mod tests {
     }
 
     #[test]
-    fn playground_test() {
-        let current_crate = load_rustdoc_from_file(Path::new("../cargo-semver-check/localdata/test_data/struct_repr_c_removed.json"));
+    fn attribute_from_string_simple_inner() {
+        let attribute = Attribute::try_from(&"#![no_std]".to_string()).unwrap();
+        assert_eq!(attribute, Attribute {
+            is_inner: true,
+            content: AttributeValue {
+                as_string: "no_std".to_string(),
+                base: "no_std".to_string(),
+                assigned_expression: None,
+                arguments: None
+            }
+        });
+    }
 
-        let current = IndexedCrate::new(&current_crate);
-        let query = r#"
-        {
-            Crate {
-                item {
-                    ... on Enum {
-                        visibility_limit @filter(op: "=", value: ["$public"])
-                        name @output
-                        attribute {
-                            is_inner @output
-                            content {
-                                base @output
-                                assigned_expression @output
-                                argument {
-                                    as_string @output
-                                }
+    #[test]
+    fn attribute_from_string_complex_outer() {
+        let attribute = Attribute::try_from(&"#[cfg_attr(feature = \"serde\", derive(Serialize, Deserialize))]".to_string()).unwrap();
+        assert_eq!(attribute, Attribute {
+            is_inner: false,
+            content: AttributeValue {
+                as_string: "cfg_attr(feature = \"serde\", derive(Serialize, Deserialize))".to_string(),
+                base: "cfg_attr".to_string(),
+                assigned_expression: None,
+                arguments: Some(vec![
+                    AttributeValue {
+                        as_string: "feature = \"serde\"".to_string(),
+                        base: "feature".to_string(),
+                        assigned_expression: Some("\"serde\"".to_string()),
+                        arguments: None
+                    },
+                    AttributeValue {
+                        as_string: " derive(Serialize, Deserialize)".to_string(),
+                        base: "derive".to_string(),
+                        assigned_expression: None,
+                        arguments: Some(vec![
+                            AttributeValue {
+                                as_string: "Serialize".to_string(),
+                                base: "Serialize".to_string(),
+                                assigned_expression: None,
+                                arguments: None
+                            },
+                            AttributeValue {
+                                as_string: " Deserialize".to_string(),
+                                base: "Deserialize".to_string(),
+                                assigned_expression: None,
+                                arguments: None
                             }
-                        }
+                        ]) 
                     }
-                }
+                ])
             }
-        }"#;
-        let mut arguments = BTreeMap::new();
-        arguments.insert("public", "public");
+        });
+    }
 
-        let schema = RustdocAdapter::schema();
-        let adapter = Rc::new(RefCell::new(RustdocAdapter::new(&current, None)));
-
-        let parsed_query = parse(&schema, query).unwrap();
-        let args = Arc::new(
-            arguments
-                .iter()
-                .map(|(k, v)| (Arc::from(k.to_string()), (*v).into()))
-                .collect(),
-        );
-        let results_iter = interpret_ir(adapter.clone(), parsed_query, args).unwrap();
-
-        let actual_results: Vec<BTreeMap<_, _>> = results_iter
-            .map(|res| res.into_iter().map(|(k, v)| (k.to_string(), v)).collect())
-            .collect();
-
-        for row in actual_results.iter() {
-            for (k, v) in row {
-                println!("{} {:?}", k, v);
-            }
-            println!("");
+    #[test]
+    fn attribute_value_from_string_custom_brackets() {
+        for as_string in ["macro{arg1,arg2}", "macro[arg1,arg2]"] {
+            let attr_val = AttributeValue::try_from(&as_string.to_string()).unwrap();
+            assert_eq!(attr_val, AttributeValue {
+                as_string: as_string.to_string(),
+                base: "macro".to_string(),
+                assigned_expression: None,
+                arguments: Some(vec![
+                    AttributeValue {
+                        as_string: "arg1".to_string(),
+                        base: "arg1".to_string(),
+                        assigned_expression: None,
+                        arguments: None
+                    },
+                    AttributeValue {
+                        as_string: "arg2".to_string(),
+                        base: "arg2".to_string(),
+                        assigned_expression: None,
+                        arguments: None                     
+                    }
+                ])
+            });
         }
     }
 
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -135,6 +135,16 @@ pub struct Attribute {
     content: AttributeValue,
 }
 
+impl Attribute {
+    fn as_string(&self) -> String {
+        format!(
+            "#{}[{}]",
+            if self.is_inner { "!" } else { "" },
+            self.content.as_string
+        )
+    }
+}
+
 impl<'a> TryFrom<&'a str> for Attribute {
     type Error = anyhow::Error;
 
@@ -571,6 +581,7 @@ fn get_impl_property(token: &Token, field_name: &str) -> FieldValue {
 fn get_attribute_property(token: &Token, field_name: &str) -> FieldValue {
     let attribute = token.as_attribute().expect("token was not an Attribute");
     match field_name {
+        "as_string" => attribute.as_string().into(),
         "is_inner" => attribute.is_inner.clone().into(),
         _ => unreachable!("Attribute property {field_name}"),
     }
@@ -1462,6 +1473,7 @@ mod tests {
                 }
             }
         );
+        assert_eq!(attribute.as_string(), "#![no_std]");
     }
 
     #[test]
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -13,8 +13,8 @@ use trustfall_core::{
 use crate::indexed_crate::IndexedCrate;
 
 use anyhow::bail;
-use regex::Regex;
 use lazy_static::lazy_static;
+use regex::Regex;
 
 #[non_exhaustive]
 pub struct RustdocAdapter<'a> {
@@ -135,7 +135,7 @@ pub struct Attribute {
     content: AttributeValue,
 }
 
-impl<'a> TryFrom<&'a str> for Attribute {    
+impl<'a> TryFrom<&'a str> for Attribute {
     type Error = anyhow::Error;
 
     fn try_from(as_string: &'a str) -> anyhow::Result<Self> {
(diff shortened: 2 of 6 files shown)


src/adapter.rs
```diff
@@ -113,10 +124,140 @@ pub enum TokenKind<'a> {
...
+                content: AttributeValue::try_from(&captures[1])?,
+            }),
+            None => match OUTER_RE.captures(as_string) {
+                Some(captures) => Ok(Attribute {
+                    is_inner: false,
+                    content: AttributeValue::try_from(&captures[1])?,
+                }),
+                None => bail!("Attribute has to be in one of the following forms: `#[...]` or `#![...]`, but found: `{}`", as_string)
+            }
+        }
```
obi1kenobi: Regex here seems a bit too complex. We're trying to do two things:
- assert that the string starts with `#[` (outer) or `#![` (inner), otherwise error
- assert that the string ends with `]`, otherwise error

This should be doable without regex, with only built-in functions — and the separation of steps would almost certainly produce better errors.

src/adapter.rs
```diff
@@ -113,10 +124,140 @@ pub enum TokenKind<'a> {
...
+    base: String,
+    assigned_expression: Option<String>,
+    arguments: Option<Vec<AttributeValue>>,
+}
+
+impl<'a> TryFrom<&'a str> for AttributeValue {
+    type Error = anyhow::Error;
+
+    // TODO: Are we sure it recognises all possible attributes? Maybe it's better
+    // to ignore errors instead of panicing?
```
obi1kenobi: How about this as an option: failure to parse the content components leaves everything as-is in `base` and leaves the assigned expression and arguments `None`.

Panics and error returns aren't great from a downstream UX perspective because they both ultimately prevent querying. If something claims it's an attribute but doesn't start with `#[` or `#![`, or doesn't end with `]`, I'm comfortable panicking because that's against [the spec](https://doc.rust-lang.org/reference/attributes.html). But if something doesn't match our ability to parse it out into components, I think it's best to let it slide without errors or panics. For example, we don't want to prevent semver-checking a crate just because an item in the crate has an unusual and unexpected attribute format.

With this in mind, maybe `TryFrom` isn't the right trait here? Would a `::new()` method would be a better option than implementing `From<&str>`?

src/adapter.rs
```diff
@@ -113,10 +124,140 @@ pub enum TokenKind<'a> {
...
+                    .unwrap();
+            static ref ARGUMENTS_RE: Regex = Regex::new(
+                format!(
+                    r"^{}[\(\[\{{](?P<arguments>.*)[\)\]\}}][[:blank:]]*$",
+                    PATH_RE_STR
+                )
+                .as_str()
+            )
+            .unwrap();
+        }
```
obi1kenobi: This is quite a bit of regex, and I suspect it's going to be difficult to understand and maintain, as well as possibly having difficult-to-predict bugs.

Would we be better off with a proper parser-based approach here instead?

src/rustdoc_schema.graphql
```diff
@@ -499,12 +499,18 @@ type Method implements Item & FunctionLike {
...
+  is_inner: Boolean!
+  content: AttributeValue!
+}
 
-  For example: "#[repr(C)]"
-  """
-  value: String!
+type AttributeValue {
+  base: String!  # the path, e.g. "cfg_attr" or "must_use"
+  as_string: String!  # the entire value represented as a string
```
obi1kenobi: Two suggestions on this:
- Let's rename `as_string` to something like `full_value` or `raw_value`. Right now, it's confusing because all the other properties are strings too, so `as_string` doesn't differentiate itself.
- Let's move `as_string` to the top of the property list. It's the simplest-to-understand property, and readers should see it first before moving to anything more sophisticated.

In general, you might notice a theme here: when we have extra degrees of freedom, let's use them in a way that optimizes for the downstream user's experience. For example, the schema parser doesn't care about property order (i.e. it's a degree of freedom) so let's lay out the properties in an order that will help the humans that read this file.

src/rustdoc_schema.graphql
```diff
@@ -499,12 +499,18 @@ type Method implements Item & FunctionLike {
...
+  assigned_expression: String  # the string form of the assigned expression, or null if there wasn't one
```
obi1kenobi: I'm inclined to suggest `assigned_value` here instead of `assigned_expression`: it would be consistent with `full_value / raw_value` above, and `expression` is kind of a "scary" word.

In general, whenever it doesn't hurt something else, it's good to use less-complex language as much as possible because it makes onboarding new users faster and easier. Naming simple ideas with complex names keeps people away from learning and using them.

src/rustdoc_schema.graphql
```diff
@@ -499,12 +499,18 @@ type Method implements Item & FunctionLike {
 A specific attribute applied to an Item.
 """
 type Attribute {
-  """
-  The textual representation of the attribute.
+  as_string: String!
+  is_inner: Boolean!
+  content: AttributeValue!
```
obi1kenobi: `content` is an edge but the others are properties. Please document these fields and keep consistency in the schema with respect to how the properties and edges are formatted (e.g. blank line between properties and edges, comments, etc.).

src/adapter.rs
```diff
@@ -1235,11 +1451,106 @@ mod tests {
 
     #[test]
     fn rustdoc_json_format_version() {
-        let current_crate = load_rustdoc_from_file(Path::new("./test_data/rustdoc_v21.json"));
+        let version = rustdoc_types::FORMAT_VERSION;
+        let current_crate =
+            load_rustdoc_from_file(Path::new(&format!("./test_data/rustdoc_v{version}.json")));
```
obi1kenobi: Won't this change cause a worse error than the original? With this change it seems like we'd get a "file not found" error when there's a version mismatch, instead of a "version mismatch" error.

src/adapter.rs
```diff
@@ -113,10 +124,140 @@ pub enum TokenKind<'a> {
...
+                            depth -= 1;
+                        } else if c == ',' && depth == 0 {
+                            arguments.push(AttributeValue::try_from(&arg_str[i..j]).unwrap());
+                            i = j + 1;
+                            only_white = true;
+                        }
+                    }
+                    if i < arg_str.len() && !only_white {
+                        arguments.push(AttributeValue::try_from(&arg_str[i..]).unwrap());
+                    }
```
obi1kenobi: I know it's very easy to slip into "IOI mode," but it's very important to not do that — especially in code that is going to be long-lived, and will be maintained for a long time by people who aren't just yourselves. This code is practically unreadable to anyone that didn't do IOI / ACM and hasn't been trained on understanding it quickly for purposes of submitting challenges on TopCoder and similar sites :)

Things you can do to improve this:
- Use more descriptive variable names than `i, j, c` etc.
- Write comments describing what is going on.
- Extract this code out of `TryFrom` and even out of `adapter.rs`, into its own file and break it up into separate, smaller functions that are well-named, individually-tested, and documented.

obi1kenobi (commented):
> The instances of `Attribute` and `AttributeValue` are owned by tokens and therefore cloned each time a new token is created. How can we improve this? Move the ownership somewhere else? Use `Rc`?

`Rc` is reasonable. Another option, perhaps best used in combination, is borrowing from the underlying strings as much as possible since almost all the operations we're doing here are slicing smaller strings out of bigger ones.

If we have a `bool` and a few `&str`, cloning is totally fine and I don't think `Rc` will be worth it. But use your judgment and do a pass on this and we can look at it together from there.

> `TryFrom` impl for `AttributeValue` panics if the given string is not in one of three recognized forms. They seem to cover all possible cases, but are we sure about that? Maybe it would be better to somehow ignore it (for example returning `AttributeValue` with only `as_string` set), so that the adapter will not fail on some attributes which are indifferent to us?

Yes, I agree that panicking on an unrecognized attribute format would be unfortunate. See the PR comments, I made a concrete suggestion there.

> Ignore white spaces or not (especially in `as_string`)? For example, when the attribute is `#[repr(C, u8)]`, the second argument will have `as_string=' u8'`, but `base='u8'`. 

I think it's fine to strip leading and trailing whitespace from strings in `as_string`. I doubt anyone would complain and want it back, since I think it would get normalized out in `rustfmt` anyway.

> Naming. I like the names of properties and edges, but I am not really sure about `AttributeValue`.

Naming is always hard :) Do you have any other ideas for names we could consider?

src/adapter.rs
mgr0dzicki: You're right, regexes here are an overkill, I will change it!

src/adapter.rs
mgr0dzicki: I agree that leaving everything in `base` if we are not able to parse it seems to be the best option, and therefore definitely `new` is better then `TryFrom`.

But then I have another questions: can we assume the code is formatted by `cargo fmt`, so there won't be leading and trailing spaces (like "    #[non_axhaustive]   ")? And what about the attributes of nested structs for example? If the struct and its attributes are indented, then does the attribute's string from rustdoc contain indentation spaces?

src/rustdoc_schema.graphql
mgr0dzicki: Fully agree! I like `raw_value` more than `full_value`, I think it better indicates that it's the value before any parsing.

src/rustdoc_schema.graphql
mgr0dzicki: I'm sorry, I should have looked over other types in the schema and seen how they are formatted, I forgot about that... Of course it will be changed

src/adapter.rs
mgr0dzicki: I'm sorry for this, the change is completely unintentional... At some moment I accidentally removed this test, but noticed it only later and wasn't able to simply undo, so I copied it... but from the wrong branch.

src/adapter.rs
mgr0dzicki: You're right, probably even me after a year wouldn't be able to understand this code easily... I'll improve it!

What about making a separate module `attributes`? Or maybe it should be a sub-module of `adapter.rs`? Definitions of Attribute and AttributeValue, parsing them and specific tests would be more then two hundred lines of code, which in fact does not depend on anything else from the crate.

mgr0dzicki:
> > The instances of `Attribute` and `AttributeValue` are owned by tokens and therefore cloned each time a new token is created. How can we improve this? Move the ownership somewhere else? Use `Rc`?
> 
> `Rc` is reasonable. Another option, perhaps best used in combination, is borrowing from the underlying strings as much as possible since almost all the operations we're doing here are slicing smaller strings out of bigger ones.
> 
> If we have a `bool` and a few `&str`, cloning is totally fine and I don't think `Rc` will be worth it. But use your judgment and do a pass on this and we can look at it together from there.

So I'll first try to borrow as much as we can without using `Rc`.

> > Naming. I like the names of properties and edges, but I am not really sure about `AttributeValue`.
> 
> Naming is always hard :) Do you have any other ideas for names we could consider?

Then I didn't, but now... I've recently found out that our `AttributeValue` is called a `meta item` (see https://doc.rust-lang.org/reference/attributes.html#meta-item-attribute-syntax), so what about naming the struct `MetaItem`? I'm a little concerned that the reference states it is "the syntax used for the Attr rule by most built-in attributes". "Most" suggests "not all", but I haven't found any examples... And even if there are any, it shouldn't be a problem (assuming we don't panic during parsing) unless we need to write a lint checking for such an attribute, right?

mgr0dzicki:
And my last question for now: what would be the best way to handle assumptions in parsing? For example, we assume the attribute is either outer `#[...]` or inner `#![...]`. Should we then
- check if the assumption holds true, and panic otherwise,
- or just assume it's true and don't care what will be returned when it isn't?

Cargo.toml
```diff
@@ -13,7 +13,7 @@ readme = "./README.md"
 [dependencies]
 trustfall_core = "0.0.4"
 rustdoc-types = "0.17.0"
-anyhow = "1.0.66"
+anyhow = "1.0.58"
```
tonowak: A version downgrade?

src/adapter.rs
obi1kenobi: I think in general we should make the fewest number of assumptions possible. So we shouldn't assume that the code is formatted in any particular way, merely that it's valid Rust.

That said, I think it's fine to trim leading/trailing whitespace in situations where that whitespace is semantically unimportant.

obi1kenobi:
> And my last question for now: what would be the best way to handle assumptions in parsing? For example, we assume the attribute is either outer #[...] or inner #![...].

I think it's fine to panic if the attribute is neither inner nor outer.

In general:
- Surprising states that are still representable in our schema should not panic. An example would be an attribute that is none of the three forms we parse (`#[foo]` / `#[foo(bar)]` / `#[foo = bar]`).
- Surprising states that have no valid representation in our schema should error or panic rather than returning invalid data. This inner vs outer attribute situation is a good example of that.

In an alternative schema design, we could have made the `is_inner: Boolean!` property allow nullability (`is_inner: Boolean`) and use the null value for these "neither inner nor outer" attributes. But this seems to hurt user ergonomics for almost everyone, for an unclear and dubious benefit to anyone.

This is why good schema design is the most difficult part of building a good Trustfall integration for a data source: it's all the little decisions about how to best model the data, capturing all the possible edge cases without making the schema overly complicated and difficult to use.

Also if choosing between errors and panics, consider what the caller's options are: is there ever a situation where they could handle an error in a way other than panicking? For example: `trustfall` adapters don't currently have an ability to handle errors that happen in the middle of a query, so if a function called by an adapter returns `Result`, the adapter will just immediately `.expect()` on it. In this case, it's better to panic immediately rather than add more complexity and more dependencies (like `anyhow`). This then has downstream effects: use `new()` instead of implementing `From / TryFrom` etc.

src/adapter.rs
obi1kenobi: Look at the `strip_prefix()` and `strip_suffix()` methods, I tend to like their ergonomics for situations like this:
https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix

mgr0dzicki:
@obi1kenobi I tried changing String's to &str's as we discussed, but I ran into lifetime errors and I don't know how to fix them. Could you have a look when you have a minute?

obi1kenobi:
The TL;DR of the issue is that `Token::as_attribute()` needs to return an `Option` of either an owned type or an `'a`-lifetime reference in order to satisfy the lifetime needs of the iterator type in the return value.

In the other `as_x()` methods, the data is owned by the rustdoc JSON and not transformed in any way, so the methods return `&'a`. However, your code's `Token::as_attribute()` and `as_attribute_value()` return a `&'_` which means "the same lifetime as `&self`" which isn't long enough.

There are two possible ways to fix this:
- The adapter already uses `&'a IndexedCrate<'a>` to own some data that is generated from rustdoc and not rustdoc directly — see the `visibility_forest` field for example. In principle, it should be possible to make the adapter maintain a cache of "this rustdoc item has this `Attribute<'a>` value" and have the lifetimes work out. But this will probably be a bit involved since it would involve mutating the adapter's contents from inside the output iterator's closure.
- If "return `&'a` data" doesn't work, we can instead make the data owned and cheaply-cloneable. You could make `TokenKind` use variants like:
```rust
    Attribute(Rc<Attribute<'a>>),
    AttributeValue(Rc<AttributeValue<'a>>),
```
and have the `Attribute` internally use `Rc<AttributeValue>` as well.

I'd probably choose the latter option, since it's a good tradeoff between implementation complexity and speed (it still borrows `&'a str` instead of owning the strings).

I'd be happy to pair-program this with you, if that would be helpful.

mgr0dzicki:
> The TL;DR of the issue is that `Token::as_attribute()` needs to return an `Option` of either an owned type or an `'a`-lifetime reference in order to satisfy the lifetime needs of the iterator type in the return value.

I finally managed to fix this by simply changing the return type of `as_attribute` from `Option<&'_ Attribute>` to `Option<&'_ Attribute<'a>>`, and similarly for `as_attribute_value`. No need to use `Rc` in the end :) Now I'm working on testing the code and making it more clear.

obi1kenobi:
> I finally managed to fix this by simply changing the return type of `as_attribute` from `Option<&'_ Attribute>` to `Option<&'_ Attribute<'a>>`, and similarly for `as_attribute_value`. No need to use `Rc` in the end :) Now I'm working on testing the code and making it more clear.

Consider what happens when cloning a vertex (`Token`), for example with this line:
https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/5/files#diff-58c62a04b04e4455b95e7916cdab35c54e6b8da91359eaa526096c88b1f1cdffR43

I'd either put an `Rc` around that vector so we don't have to clone it every time a vertex needs cloning, or `Rc` the whole `AttributeValue` variant in `TokenKind`. References and options are cheap to clone, but vectors aren't necessarily so.

mgr0dzicki:
> > I finally managed to fix this by simply changing the return type of `as_attribute` from `Option<&'_ Attribute>` to `Option<&'_ Attribute<'a>>`, and similarly for `as_attribute_value`. No need to use `Rc` in the end :) Now I'm working on testing the code and making it more clear.
> 
> Consider what happens when cloning a vertex (`Token`), for example with this line: https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/5/files#diff-58c62a04b04e4455b95e7916cdab35c54e6b8da91359eaa526096c88b1f1cdffR43
> 
> I'd either put an `Rc` around that vector so we don't have to clone it every time a vertex needs cloning, or `Rc` the whole `AttributeValue` variant in `TokenKind`. References and options are cheap to clone, but vectors aren't necessarily so.

You're right, I forgot about that copying... Now I store `AttributeValue`'s using `Rc`, is this what you had in mind?

Cargo.toml
mgr0dzicki: Not used anymore ;)

src/adapter.rs
```diff
@@ -1192,6 +1223,55 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
...
+                    let neighbors: Box<dyn Iterator<Item = Self::DataToken> + 'a> =
+                        match &ctx.current_token {
+                            None => Box::new(std::iter::empty()),
+                            Some(token) => {
+                                let origin = token.origin;
+
+                                let attr_value = token
+                                    .as_attribute_value()
+                                    .expect("token was not an AttributeValue");
+                                if let Some(arguments) = attr_value.arguments.clone() {
```
obi1kenobi: This `.clone()` is still cloning an entire vector, even though the vector contains `Rc` values. Consider making it an `Rc<Vec<...>>` type, since we'll never change the contents of the vector?

I think this might be more important than wrapping `AttributeValue` in `Rc`, since if the vector is ref-counted, then then entirety of a non-refcounted `AttributeValue` can still be cloned in just 32 bytes: it's only 4 pointers.

src/attributes.rs
```diff
@@ -0,0 +1,235 @@
...
+                is_inner: false,
+                content: Rc::new(AttributeValue::new(raw_content)),
+            }
+        } else if let Some(raw_content) = raw_without_closing.strip_prefix("#![") {
+            Attribute {
+                is_inner: true,
+                content: Rc::new(AttributeValue::new(raw_content)),
+            }
+        } else {
+            panic!("Attribute has to start either with `#[` or `#![`.");
```
obi1kenobi: Consider what would happen if this `panic!()` got hit: we'd see this static string in the error message, and nothing else.

So the first thing we'd then do is edit the `panic!()` call to include additional context, like what the `raw` value that triggered this was. And then we'd try to replicate the crash to see the error with that additional context, so we can debug it further.

We can save a step here: just include all the necessary context immediately, without waiting for the panic to be hit in the real world first. This is a way to reduce future debugging time, without significant upfront cost, and is highly effective in my experience.

There are a few more instances of this kind of situation throughout this PR, and I'd encourage you to self-review it to practice noticing them as well as fixing them.

src/attributes.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
```
obi1kenobi: Any particular reason this isn't also `Eq`?

obi1kenobi (commented):
I know you are still polishing the code, especially around the parsing, so I didn't look at the parsing parts particularly closely.

Just a few other general observations and suggestions for improvement in the meantime.

src/adapter.rs
mgr0dzicki: Could you describe the approach with `Rc<Vec<...>>` in more detail? I tried changing `Vec<...>` to `Rc<Vec<...>>` but then
```
if let Some(arguments) = attr_value.arguments.clone() {
    Box::new(arguments.iter().map(move |argument| {
        origin.make_attribute_value_token(argument.clone())
    }))
}
```
causes
```
error[E0597]: `arguments` does not live long enough
```
and I don't know how to fix it without cloning the vector (which might be consumed by `into_iter`) or collecting the resulting iterator (but this requires creating a new vector too...)

src/attributes.rs
mgr0dzicki: No, I just didn't derive it, because I need only PartialEq. Is it better, in general, to derive both if possible?

src/attributes.rs
mgr0dzicki: You're definitely right, I'll change it!

mgr0dzicki:
> Then I didn't, but now... I've recently found out that our `AttributeValue` is called a `meta item` (see https://doc.rust-lang.org/reference/attributes.html#meta-item-attribute-syntax), so what about naming the struct `MetaItem`? I'm a little concerned that the reference states it is "the syntax used for the Attr rule by most built-in attributes". "Most" suggests "not all", but I haven't found any examples... And even if there are any, it shouldn't be a problem (assuming we don't panic during parsing) unless we need to write a lint checking for such an attribute, right?

And what do you think about renaming `AttributeValue` to `MetaItem`, as it is called in the reference? ;)

src/attributes.rs
obi1kenobi: Yes, if the type is able to satisfy `Eq` then it's best to derive it. There's no downside, only upside.

You probably have heard of "[Chesterton's Fence](https://wiki.lesswrong.com/wiki/Chesterton%27s_Fence)": the principle that if something looks strange, it's likely that way for a reason even if the reason may be non-obvious, and that this reason must be understood before the system is modified. In this situation, we have a corollary of Chesterton's Fence: don't design things in an unusual way when there's no reason for that. If I was reading some code and saw a type that is `PartialEq` but isn't `Eq`, I'd assume that's for some non-obvious reason and would try to figure out why `Eq` might not be possible to satisfy.

This corollary of Chesterton's Fence is sometimes also called the Principle of Least Astonishment: https://en.wikipedia.org/wiki/Principle_of_least_astonishment

obi1kenobi:
Sorry, I missed this. Thanks for the reminder.

> Then I didn't, but now... I've recently found out that our `AttributeValue` is called a `meta item` (see https://doc.rust-lang.org/reference/attributes.html#meta-item-attribute-syntax), so what about naming the struct `MetaItem`?

`MetaItem` seems a bit vague for our schema: seeing `type MetaItem` in the schema would be confusing and wouldn't really give the reader any impression as to where these vertices might appear.

I'd be open to naming it `AttributeMetaItem` and linking to the reference in the doc comment in the schema, if you think that's appropriate. I might also be open to leaving it called `AttributeValue` with a doc comment note that it corresponds to the `MetaItem` type in the reference together with the reference link.

What do you think is best?

> I'm a little concerned that the reference states it is "the syntax used for the Attr rule by most built-in attributes". "Most" suggests "not all", but I haven't found any examples... And even if there are any, it shouldn't be a problem (assuming we don't panic during parsing) unless we need to write a lint checking for such an attribute, right?

The way I see it, we have a few broad guiding principles that are desirable here:
- Our code shouldn't panic during parsing of any attribute syntax that rustc will accept without errors. If rustc would error, panicking is allowed but not guaranteed.
- Our code shouldn't assume that the input is formatted by rustfmt. It should produce identical vertex configurations regardless of formatting.
- The attribute syntax used by "most" attributes is parsed with more detail (separated-out `AttributeMetaItem` vertices etc.), whereas non-standard syntax is currently not parsed beyond being available in the appropriate `as_string` properties. In the future, we may add more in-depth parsing of such alternative syntax.

We should strive for these objectives, but we may not be able to achieve all of them simultaneously. In such a case, we should clearly document where we fall short, so as to minimize the risk of violating the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

src/adapter.rs
obi1kenobi: After digging into this quite a bit myself, I have to admit it seems like this is not an optimization that's worth implementing right now, and we're better off not using an `Rc<Vec<...>>` and instead just cloning the `Vec` each time the edge is accessed as you had already implemented.

The "`arguments` does not live long enough" error is because the iterator type constructed by `.iter().map(...)` is borrowing from `arguments`, which is owned because of the `.clone()` but then dropped at the end of the function i.e. while its iterator is still live. We need to pass ownership of `arguments` into the iterator to avoid this problem.

As you've already found, `.into_iter()` doesn't work because we can't consume the vector underlying the `Rc`. Since this is because `.into_iter()` wants to output owned versions of the vector's elements, I was hoping we'd be able to avoid that problem by cloning the vector's elements before returning them — kind of like a `.into_iter().cloned()` operation but combined into one wrapping struct that takes ownership of the `Rc<Vec<...>>`.

However, after taking a look at how `.into_iter()` is [implemented for](https://doc.rust-lang.org/src/alloc/vec/into_iter.rs.html#145) `Vec`, it seems like that would involve a fair bit of `unsafe` code. The risk-vs-reward here is not worth going down that path, as it would add significant complexity and risk of bugs, for a relatively small perf gain in the big picture: many lints won't read attributes i.e. won't hit this code path, and most attributes only have a small number of `AttributeValue` components so the vectors would be small and relatively fast to clone.

Sorry to lead you down the wrong path here. You had the right approach already.

src/attributes.rs
mgr0dzicki: It should be better now :)

mgr0dzicki:
> I'd be open to naming it `AttributeMetaItem` and linking to the reference in the doc comment in the schema, if you think that's appropriate. I might also be open to leaving it called `AttributeValue` with a doc comment note that it corresponds to the `MetaItem` type in the reference together with the reference link.
> 
> What do you think is best?

I really like `AttributeMetaItem`, because it both emphasizes the use of the type (`Attribute`) and references the documentation (`MetaItem`). I changed the name. Do you think that now properties `raw_value` and `assigned_value` should change their names too (for example to `raw_meta_item`, `assigned_meta_item` or `raw_item`, `assigned_item`)?

obi1kenobi:
> Do you think that now properties `raw_value` and `assigned_value` should change their names too (for example to `raw_meta_item`, `assigned_meta_item` or `raw_item`, `assigned_item`)?

I'd be okay with that, `raw_item` and `assigned_item` sound fine: clear yet not too long. It sounds like you think they should change too? Feel free to just make the change you think is right in the PR without asking for permission first -- you have good judgment and frequently we can go faster by just skipping a step :)

mgr0dzicki:
> I'd be okay with that, `raw_item` and `assigned_item` sound fine: clear yet not too long. It sounds like you think they should change too? Feel free to just make the change you think is right in the PR without asking for permission first -- you have good judgment and frequently we can go faster by just skipping a step :)

Great, I did the change :). I thought it would also require modifications in https://github.com/obi1kenobi/cargo-semver-checks/pull/171, but fortunately those properties are not being used there yet.

Now this PR is finally ready for review :)

src/rustdoc_schema.graphql
```diff
@@ -517,11 +517,62 @@ A specific attribute applied to an Item.
...
 type Attribute {
   """
-  The textual representation of the attribute.
+  String representation of the attribute as it is found in the code.
 
-  For example: "#[repr(C)]"
+  For example: `#[non_exhaustive]`
   """
-  value: String!
+  as_string: String!
```
obi1kenobi: What do you think about renaming `as_string` to something like `raw_attribute`, analogously to how `raw_item` on `AttributeMetaItem` is the raw string representation for that type?

src/rustdoc_schema.graphql
```diff
@@ -517,11 +517,62 @@ A specific attribute applied to an Item.
...
+
+  # edges
+
+  """
+  Inner meta items if the meta item is in the form `SimplePath(MetaSeq)`.
+
+  For example: `[AttributeMetaItem::new("Debug"), AttributeMetaItem::new("Clone")]`
+  for `derive(Debug, Clone)`
+  """
+  argument: [AttributeMetaItem!]
```
obi1kenobi, line 575: These docs are all 💯

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+            raw_item: raw_trimmed,
+            base: raw_trimmed,
+            assigned_item: None,
+            arguments: None,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
obi1kenobi, line 169: The test cases are great!

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn new(raw: &'a str) -> Self {
+        let raw_trimmed = raw.trim();
+        let raw_without_closing = raw_trimmed.strip_suffix(']').unwrap_or_else(|| {
+            panic!(
+                concat!(
+                    "String `{}` cannot be parsed as an attribute ",
+                    "because it is not closed with a square bracket."
+                ),
+                raw_trimmed
+            )
```
obi1kenobi: Here's a slightly more idiomatic way to do what I believe is the same thing:
```suggestion
            panic!(
                "\
String `{raw_trimmed}` cannot be parsed as an attribute \
because it is not closed with a square bracket."
            )
```

Newlines prefixed with `\` are ignored. Also, either the current or the upcoming version of clippy includes a lint that wants all formatting done inline when possible (like `{raw_trimmed}` instead of `"{}", raw_trimmed` above). We'll need to migrate all uses in the codebase to that style when that lint becomes active by default, so let's try to start following it now so don't have to do as much migrating work later.

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Tries to parse `raw` as a comma-separated sequence of `AttributeMetaItem`'s
+    /// wrapped in parentheses, square brackets or curly brackets.
+    fn slice_arguments(raw: &'a str) -> Option<Vec<Rc<AttributeMetaItem<'a>>>> {
+        let is_left = |c| c == '(' || c == '[' || c == '{';
+        let is_right = |c| c == ')' || c == ']' || c == '}';
+        let matching_right = |c| match c {
+            '(' => ')',
+            '[' => ']',
+            '{' => '}',
+            _ => unreachable!("Tried to find matching right bracket for {}.", c),
```
obi1kenobi:
```suggestion
            _ => unreachable!("Tried to find matching right bracket for {c}."),
```
Also because of that clippy lint I mentioned earlier. Try to replace all the other similar uses as well.

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        let raw_trimmed = raw.trim();
+        let first_char = raw_trimmed.chars().next()?;
+        let raw_meta_seq = raw_trimmed
+            .strip_prefix(is_left)?
+            .strip_suffix(|c| c == matching_right(first_char))?
+            .trim();
+
+        let mut i = 0; // index of the first character after the last comma
+        let mut bracket_depth = 0; // number of currently opened brackets
```
obi1kenobi: Consider keeping a stack (`Vec<char>`, since pushing and popping off the end are cheap) of brackets instead of just the depth, so you can make sure that you are tracking the depth correctly and closing the correct bracket scopes, or panic if an unexpected bracket is encountered. Then `bracket_depth` is just the length of that stack.

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+        let mut arguments: Vec<Rc<AttributeMetaItem>> = Vec::new(); // meta items constructed so far
+
+        for (j, c) in raw_meta_seq.chars().enumerate() {
+            if is_left(c) {
+                bracket_depth += 1;
+            } else if is_right(c) {
+                bracket_depth -= 1;
+            } else if c == ',' {
```
obi1kenobi: Can you add a test case that includes some parens and square brackets in a string literal, to make sure this parsing logic works correctly? I'm a bit worried about things like `#[foo("parens ) and brackets ] and curlies } here")]` being mis-parsed. Also similarly with open ones, or a mix of open and closed, and also with escaped quotes added in.

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+        };
+
+        let raw_trimmed = raw.trim();
+        let first_char = raw_trimmed.chars().next()?;
+        let raw_meta_seq = raw_trimmed
+            .strip_prefix(is_left)?
+            .strip_suffix(|c| c == matching_right(first_char))?
+            .trim();
+
+        let mut i = 0; // index of the first character after the last comma
```
obi1kenobi: Consider naming this variable `index_after_last_comma`? Then the comment won't be necessary, and the later code that uses `i` will be more obvious as well.

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+        // If the last comma was not a trailing one, there is still one meta item left.
+        if i < raw_meta_seq.len() {
+            arguments.push(Rc::new(AttributeMetaItem::new(&raw_meta_seq[i..])));
+        }
+
+        Some(arguments)
+    }
+
+    pub fn new(raw: &'a str) -> Self {
+        let simple_path_char = |c: char| c.is_alphanumeric() || c == '_' || c == ':';
```
obi1kenobi: Can you double-check the Rust identifier naming rules to see if there's anything not covered here? It would be quite embarrassing if our code failed e.g. on some internationalization-focused crate when English isn't the first language of anyone that was involved in any of this code 😅

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+        let raw_trimmed = raw.trim();
+
+        if let Some(path_end) = raw_trimmed.find(|c| !simple_path_char(c)) {
+            let simple_path = &raw_trimmed[0..path_end];
+            let attr_input = &raw_trimmed[path_end..];
```
obi1kenobi: Can you double-check whether this is all in char-index coordinates or in byte coordinates? Not all chars in UTF-8 are one byte, and I'm worried about mixing up units.

Perhaps also consider adding some test cases that use UTF-8 chars that are valid in identifiers but are not one byte long in UTF-8, to make sure this all works correctly.

src/attributes.rs
```diff
@@ -0,0 +1,282 @@
...
+        let raw_meta_seq = raw_trimmed
+            .strip_prefix(is_left)?
+            .strip_suffix(|c| c == matching_right(first_char))?
+            .trim();
+
+        let mut i = 0; // index of the first character after the last comma
+        let mut bracket_depth = 0; // number of currently opened brackets
+        let mut arguments: Vec<Rc<AttributeMetaItem>> = Vec::new(); // meta items constructed so far
+
+        for (j, c) in raw_meta_seq.chars().enumerate() {
```
obi1kenobi: I think this is wrong in a subtle way: `j` here is in char-index coordinates, but the slicing on line 92 I believe is in byte offset coordinates. Not all chars in UTF-8 are one byte long, so it would be good to add a test case with chars that aren't one byte long to check whether this is indeed correct.

If you need byte offset coordinates, use `char_indices()` instead of `chars().enumerate()`.

obi1kenobi (commented):
Overall, the code looks much, much cleaner — well done! I particularly love the attention you've paid to the documentation in the schema and the organization of the code.

Since the code was so well-organized, I understood it a lot better, and have much more substantive feedback to offer than before. I think I spotted a couple of edge cases that we should probably add test cases for, and flagged a couple of other small things to tweak as well.

I think all those should be fairly easy things to do, and I'm excited to get this merged! 🎉

src/rustdoc_schema.graphql
mgr0dzicki: Good idea, I changed it to `raw_attribute`.

src/attributes.rs
mgr0dzicki: I used `concat!` mainly because strings without indentation seemed a bit ugly to me. But if this is the more idiomatic way, then there is no reason not to follow it.

src/attributes.rs
mgr0dzicki: This is certainly a problem, I'll start from fixing it.

src/attributes.rs
mgr0dzicki: I changed it to the safer version - since we expect that after the identifier there is either a left bracket or an assignment operator (possibly followed by white spaces), then we might proceed the input until we find one of these characters.

src/attributes.rs
mgr0dzicki: Should be fixed now!

src/attributes.rs
mgr0dzicki: Is encountering an unexpected bracket a good reason to panic? Or maybe it would be better to just return the `AttributeMetaItem` with `base=raw_item` and `assigned_item=arguments=None` as we discussed earlier? It is still possible that there is some rare syntax of attributes that is not covered by this schema.

obi1kenobi: This is a tricky question, and in the end comes down to a judgment call.

Ideally, we don't want either panics nor incompletely-parsed data with `None` values. We'll want to proactively screen crates for these unusual pieces of syntax, and Tomasz is working on a way to run `cargo-semver-checks` on all the top crates on crates.io.

A panic would make it easier to spot one of those strange pieces of syntax, since the program would crash. But if we don't panic, we could still write a query that looks for e.g. `[` or `(` or `{` characters inside `base` and find those instances as well. And not panicking means we won't immediately blow up downstream code: for example, `cargo-semver-checks` can continue checking with all the other lints, and may merely suffer some false-negatives where attributes that couldn't be parsed were hiding semver issues. I don't see another argument in favor of panicking instead.

So I'd argue in favor of not panicking.

src/attributes.rs
```diff
@@ -0,0 +1,396 @@
...
+                        })
+                    ])
+                })
+            }
+        )
+    }
+
+    #[test]
+    fn attribute_raw_identifier() {
+        let attribute = Attribute::new("#[r#derive(Debug)]");
```
obi1kenobi, line 302: Nice catch! I forgot this was an allowed syntax.

src/attributes.rs
mgr0dzicki: I definitely agree with you. The only case for panicking now is when the type of attribute cannot be recognized, which I think is reasonable. 

I changed `bracket_depth` to `Vec<char>` as you suggested.

src/attributes.rs
```diff
@@ -94,12 +94,18 @@ impl<'a> AttributeMetaItem<'a> {
...
+                    brackets.push(c);
                 } else if Self::is_right_bracket(c) {
-                    bracket_depth -= 1;
+                    if let Some(top_left) = brackets.pop() {
+                        if Self::matching_right_bracket(top_left) != c {
+                            return None;
+                        }
+                    } else {
+                        return None;
+                    }
```
obi1kenobi, line 107: A comment explaining the semantic meaning of `return None` would help explain what's going on here.
```suggestion
                    // If the brackets don't match in any way, give up on parsing 
                    // individual arguments since we don't understand the format.
                    if let Some(top_left) = brackets.pop() {
                        if Self::matching_right_bracket(top_left) != c {
                            return None;
                        }
                    } else {
                        return None;
                    }
```

obi1kenobi (approved):

src/attributes.rs
```diff
@@ -0,0 +1,404 @@
...
+        for (j, c) in raw_meta_seq.char_indices() {
+            if c == '"' && !previous_is_escape {
+                inside_string_literal = !inside_string_literal;
+            }
+
+            if !inside_string_literal {
+                if Self::is_left_bracket(c) {
+                    brackets.push(c);
+                } else if Self::is_right_bracket(c) {
+                    // If the brackets don't match in any way, give up on parsing
```
obi1kenobi: Fixing the lint I added.
```suggestion
                    // If the brackets don't match in any way, give up on parsing
```

obi1kenobi (approved):
Looks great! @mgr0dzicki do you want to open the equivalent PRs to the other rustdoc versions' branches as well?

mgr0dzicki:
> Looks great! @mgr0dzicki do you want to open the equivalent PRs to the other rustdoc versions' branches as well?

So happy it's finally ready and merged! :) Sure, I'll do it tomorrow. Is it as simple as moving all the commits to other branches? Or what might need to be changed for other versions of rustdoc?

obi1kenobi:
It should be just a cherry-pick of the merged (squashed) commit over to the
other branches, and resolving any minor merge conflicts that might happen.

If you haven't used `git cherry-pick` yet, it's very easy: just move to the
branch where you want the new commit, then run that command with the gitsha
of the commit you want as a positional argument.

PR #19, opened 2022-12-30
Add unsafe parameter to trait

staniewzki:
Signed-off-by: Michał Staniewski <m.staniewzki@gmail.com>

### DIFF
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -509,6 +509,14 @@ fn get_raw_type_property(token: &Token, field_name: &str) -> FieldValue {
     }
 }
 
+fn get_trait_property(token: &Token, field_name: &str) -> FieldValue {
+    let trait_token = token.as_trait().expect("token was not a Trait");
+    match field_name {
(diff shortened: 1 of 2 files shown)


src/adapter.rs
```diff
@@ -509,6 +509,14 @@ fn get_raw_type_property(token: &Token, field_name: &str) -> FieldValue {
     }
 }
 
+fn get_trait_property(token: &Token, field_name: &str) -> FieldValue {
+    let trait_token = token.as_trait().expect("token was not a Trait");
+    match field_name {
+        "unsafe" => trait_token.is_unsafe.into(),
```
obi1kenobi, line 515: Consider adding `auto` as well, while we're at it. We don't have a use case for it right now, but it doesn't cost anything to add, isn't likely to be removed from rustdoc JSON anytime soon, and it might be useful in the future.

obi1kenobi (approved):
Looks good! I'll hold off on merging for a bit to let you decide if you'd like to add `auto` (for auto-traits) as well.

PR #24, opened 2023-01-02
Fix rustdoc trait incompatibility

tonowak:
Fixes https://github.com/obi1kenobi/cargo-semver-checks/issues/243.

### DIFF
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1163,37 +1163,26 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                             token.as_impl().expect("not an Impl token");
 
                                         if let Some(path) = &impl_token.trait_ {
-                                            if let Some(item) = item_index.get(&path.id) {
+                                            // When the implemented trait is from the same crate
+                                            // as its definition, the trait should be present
+                                            // in `item_index`. Otherwise, the
+                                            // `rustdoc_types::Trait` is not in this rustdoc. 
+                                            // The needed foreign traits for lints are hand-written 
+                                            // in `dummy_trait_items`.
+                                            // See the comments in `src/indexed_crate.rs`
+                                            // for details.
+                                            let found_item = item_index
+                                                .get(&path.id)
+                                                .or(current_crate.dummy_trait_items.get(&path.id))
+                                                .or(previous_crate.and_then(|crate_| {
+                                                    crate_.dummy_trait_items.get(&path.id)
+                                                }));
+                                            if let Some(item) = found_item {
                                                 Box::new(std::iter::once(
                                                     origin.make_implemented_trait_token(path, item),
                                                 ))
                                             } else {
-                                                let item: Item = Item {
-                                                    id: path.id.clone(),
-                                                    crate_id: 0,
-                                                    name: Some(path.name.clone()),
-                                                    span: None,
-                                                    visibility: rustdoc_types::Visibility::Public,
-                                                    docs: None,
-                                                    links: std::collections::HashMap::new(),
-                                                    attrs: Vec::new(),
-                                                    deprecation: None,
-                                                    inner: rustdoc_types::ItemEnum::Trait(Trait {
-                                                        is_auto: true,
-                                                        is_unsafe: false,
-                                                        items: Vec::new(),
-                                                        generics: rustdoc_types::Generics {
-                                                            params: Vec::new(),
-                                                            where_predicates: Vec::new(),
-                                                        },
-                                                        bounds: Vec::new(),
-                                                        implementations: Vec::new(),
-                                                    }),
-                                                };
-                                                Box::new(std::iter::once(
-                                                    origin
-                                                        .make_implemented_trait_token(path, &item),
-                                                ))
+                                                Box::new(std::iter::empty())
                                             }
                                         } else {
                                             Box::new(std::iter::empty())
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1166,17 +1166,21 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                             // When the implemented trait is from the same crate
                                             // as its definition, the trait should be present
                                             // in `item_index`. Otherwise, the
-                                            // `rustdoc_types::Trait` is not in this rustdoc. 
-                                            // The needed foreign traits for lints are hand-written 
+                                            // `rustdoc_types::Trait` is not in this rustdoc.
+                                            // The needed foreign traits for lints are hand-written
                                             // in `dummy_trait_items`.
                                             // See the comments in `src/indexed_crate.rs`
                                             // for details.
                                             let found_item = item_index
                                                 .get(&path.id)
-                                                .or_else(|| current_crate.dummy_trait_items.get(&path.id))
-                                                .or_else(|| previous_crate.and_then(|crate_| {
-                                                    crate_.dummy_trait_items.get(&path.id)
-                                                }));
+                                                .or_else(|| {
+                                                    current_crate.dummy_trait_items.get(&path.id)
+                                                })
+                                                .or_else(|| {
+                                                    previous_crate.and_then(|crate_| {
+                                                        crate_.dummy_trait_items.get(&path.id)
+                                                    })
+                                                });
                                             if let Some(item) = found_item {
                                                 Box::new(std::iter::once(
                                                     origin.make_implemented_trait_token(path, item),
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1164,13 +1164,13 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
 
                                         if let Some(path) = &impl_token.trait_ {
                                             // When the implemented trait is from the same crate
-                                            // as its definition, the trait should be present
+                                            // as its definition, the trait is expected to be present
                                             // in `item_index`. Otherwise, the
-                                            // `rustdoc_types::Trait` is not in this rustdoc.
-                                            // The needed foreign traits for lints are hand-written
-                                            // in `manually_inlined_builtin_traits`.
-                                            // See the comments in `src/indexed_crate.rs`
-                                            // for details.
+                                            // `rustdoc_types::Trait` is not in this rustdoc,
+                                            // even if the trait is part of Rust `core` or `std`.
+                                            // As a temporary workaround, some common
+                                            // Rust built-in traits are manually "inlined"
+                                            // with items stored in `manually_inlined_builtin_traits`.
                                             let found_item = item_index
                                                 .get(&path.id)
                                                 .or_else(|| {
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1173,8 +1173,8 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                             // for details.
                                             let found_item = item_index
                                                 .get(&path.id)
-                                                .or(current_crate.dummy_trait_items.get(&path.id))
-                                                .or(previous_crate.and_then(|crate_| {
+                                                .or_else(|| current_crate.dummy_trait_items.get(&path.id))
+                                                .or_else(|| previous_crate.and_then(|crate_| {
                                                     crate_.dummy_trait_items.get(&path.id)
                                                 }));
                                             if let Some(item) = found_item {
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1174,12 +1174,15 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                             let found_item = item_index
                                                 .get(&path.id)
                                                 .or_else(|| {
-                                                    current_crate.manually_inlined_builtin_traits.get(&path.id)
-                                                })
-                                                .or_else(|| {
-                                                    previous_crate.and_then(|crate_| {
-                                                        crate_.manually_inlined_builtin_traits.get(&path.id)
-                                                    })
+                                                    let manually_inlined_builtin_traits = match origin {
+                                                        Origin::CurrentCrate => &current_crate.manually_inlined_builtin_traits,
+                                                        Origin::PreviousCrate => {
+                                                            &previous_crate
+                                                                .expect("no previous crate provided")
+                                                                .manually_inlined_builtin_traits
+                                                        }
+                                                    };
+                                                    manually_inlined_builtin_traits.get(&path.id)
                                                 });
                                             if let Some(item) = found_item {
                                                 Box::new(std::iter::once(
--- a/src/adapter.rs
+++ b/src/adapter.rs
@@ -1168,17 +1168,17 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                             // in `item_index`. Otherwise, the
                                             // `rustdoc_types::Trait` is not in this rustdoc.
                                             // The needed foreign traits for lints are hand-written
-                                            // in `dummy_trait_items`.
+                                            // in `manually_inlined_builtin_traits`.
                                             // See the comments in `src/indexed_crate.rs`
                                             // for details.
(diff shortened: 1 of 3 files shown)


src/adapter.rs
```diff
@@ -1168,7 +1168,32 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
...
+                                                            params: Vec::new(),
+                                                            where_predicates: Vec::new(),
+                                                        },
+                                                        bounds: Vec::new(),
+                                                        implementations: Vec::new(),
+                                                    }),
+                                                };
+                                                Box::new(std::iter::once(
+                                                    origin
+                                                        .make_implemented_trait_token(path, &item),
```
obi1kenobi: Try this, as a very temporary hack:
```suggestion
                                                        .make_implemented_trait_token(path, Box::leak(Box::new(item))),
```

src/adapter.rs
```diff
@@ -1163,7 +1163,21 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                                             token.as_impl().expect("not an Impl token");
 
                                         if let Some(path) = &impl_token.trait_ {
-                                            if let Some(item) = item_index.get(&path.id) {
+                                            // When the implemented trait is from the same crate
+                                            // as its definition, the trait should be present
+                                            // in `item_index`. Otherwise, the
+                                            // `rustdoc_types::Trait` is not in this rustdoc. 
+                                            // The needed foreign traits for lints are hand-written 
+                                            // in `dummy_trait_items`.
```
tonowak: What is the intention behind `trustfall-rustdoc`? Is it a tool specifically for `cargo-semver-checks`? If yes, is it OK to explain that a fix was made only for the lints to work? If no, then I guess I shouldn't mention about the existence of lints?

Nevertheless, I don't really like this comment, but I can't find a better one. Can you phrase it better @obi1kenobi ?

src/indexed_crate.rs
```diff
@@ -148,3 +156,73 @@ fn collect_public_items<'a>(
...
+        .map(|item| &item.inner)
+        .filter_map(|item_enum| match item_enum {
+            rustdoc_types::ItemEnum::Impl(impl_) => Some(impl_),
+            _ => None,
+        })
+        .filter_map(|impl_| impl_.trait_.as_ref());
+
+    // Limiting the creation of dummy traits to only those that are used by the lints.
+    // There are other foreign traits and it is not obvious how the dummy traits
+    // should look like for them.
```
tonowak: But then, what happens when a user derives a trait not from the `std` crates, but from another foreign crate? The `derive_trait_impl_removed` lint would not work.

src/indexed_crate.rs
```diff
@@ -148,3 +156,73 @@ fn collect_public_items<'a>(
...
+    .into_iter()
+    .flatten()
+    .collect();
+    let paths = paths.filter(|path| trait_names.contains(&path.name.as_str()));
+
+    paths
+        .map(|path| {
+            let dummy_item: Item = Item {
+                id: path.id.clone(),
+                crate_id: 0,
```
tonowak: Is it safe to assign `0` to `crate_id`? The `crate_id` is present in the schema after all. Is there a way to somehow know what should be the value of `crate_id`?

src/adapter.rs
```diff
@@ -1163,7 +1163,21 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
...
+                                            let found_item = item_index
+                                                .get(&path.id)
+                                                .or(current_crate.dummy_trait_items.get(&path.id))
+                                                .or(previous_crate.and_then(|crate_| {
+                                                    crate_.dummy_trait_items.get(&path.id)
+                                                }));
+                                            if let Some(item) = found_item {
                                                 Box::new(std::iter::once(
                                                     origin.make_implemented_trait_token(path, item),
                                                 ))
```
tonowak, line 1190: Should the `trait: Trait!` in `ImplementedTrait` in the schema be replaced with `trait: Trait` now that there is no guarantee that a `Trait` exists?

src/adapter.rs
obi1kenobi: Currently `cargo-semver-checks` is the only thing that uses `trustfall-rustdoc`, but I'd prefer that to not be a long-term requirement. But I think with a suitable explanation, what we're doing will be obviously preferable for any consumer of this crate.

src/indexed_crate.rs
```diff
@@ -8,15 +8,23 @@ pub struct IndexedCrate<'a> {
...
     pub(crate) visibility_forest: HashMap<&'a Id, Vec<&'a Id>>,
+
+    /// As of https://github.com/rust-lang/rust/pull/105182
+    /// the `rustdoc_types::Trait`s from foreign/outside crates
+    /// (in particular, traits like `Debug`, `Send` or `Eq`) are no longer present in `inner`.
+    /// Unfortunately, the current schema requires having `rustdoc_types::Trait` for each
+    /// trait implementation.
+    /// The `dummy_trait_items` tries to fix this problem by creating a hand-written `Trait`
+    /// for each of the std traits the user can use.
+    pub(crate) dummy_trait_items: HashMap<Id, Item>,
```
obi1kenobi: I generally dislike `dummy`-based phrasing for a variety of reasons, among which is the fact that it isn't very descriptive and its ability to communicate its purpose hinges on a lot of surrounding context. In practice that means that new contributor onboarding is more difficult than it needs to be, for both the new contributor (lots of mandatory context to learn) and for the person helping them onboard (lots of context to explain).

In general, good naming is an extremely valuable skill to practice. It makes a huge difference when programming as part of a team, which is almost all of the time.

Try `manually_inlined_builtin_traits` as a name instead?

src/indexed_crate.rs
```diff
@@ -8,15 +8,23 @@ pub struct IndexedCrate<'a> {
...
     // For an Id, give the list of item Ids under which it is publicly visible.
     pub(crate) visibility_forest: HashMap<&'a Id, Vec<&'a Id>>,
+
+    /// As of https://github.com/rust-lang/rust/pull/105182
+    /// the `rustdoc_types::Trait`s from foreign/outside crates
+    /// (in particular, traits like `Debug`, `Send` or `Eq`) are no longer present in `inner`.
+    /// Unfortunately, the current schema requires having `rustdoc_types::Trait` for each
+    /// trait implementation.
+    /// The `dummy_trait_items` tries to fix this problem by creating a hand-written `Trait`
+    /// for each of the std traits the user can use.
```
obi1kenobi: To make this easier to understand, lead with the impact, then add the context. If someone sees this comment, we want them to be able to quickly assess whether they might be affected by it or not.

What do you think of this?
```suggestion
    /// Trait items defined in external crates are not present in the `inner: &Crate` field,
    /// even if they are implemented by a type in that crate. This also includes
    /// Rust's built-in traits like `Debug, Send, Eq` etc.
    ///
    /// This change is approximately as of rustdoc v23,
    /// in https://github.com/rust-lang/rust/pull/105182
    ///
    /// As a temporary workaround, we manually create the trait items
    /// for the most common Rust built-in traits and link to those items
    /// as if they were still part of the rustdoc JSON file.
    ///
    /// A more complete future solution may generate multiple crates' rustdoc JSON
    /// and link to the external crate's trait items as necessary.
```

obi1kenobi: For extra bonus points: I noticed that `IndexedCrate` is a pub type without a doc comment, and it would be good to have something substantially similar to what I wrote here documented on the public type itself. The change we're making affects which traits are shown as implemented, so it would be good to have it documented prominently on the type itself rather than on a `pub(crate)` (i.e. private) field.

src/indexed_crate.rs
```diff
@@ -148,3 +156,73 @@ fn collect_public_items<'a>(
...
+        "Copy",
+        "PartialOrd",
+        "Ord",
+        "PartialEq",
+        "Eq",
+        "Hash",
+        "Default",
+    ];
+    let auto_traits = ["Send", "Sync", "Unpin", "RefUnwindSafe", "UnwindSafe"];
+    let other_traits = ["Sized"];
```
obi1kenobi: Lift these into a `const` (but not `pub const` -- not public API)? It would be nice to have top-level items that we can refer to in documentation that make it easy to see which traits we manually include and pretend are in the rustdoc JSON file.

src/indexed_crate.rs
```diff
@@ -148,3 +156,73 @@ fn collect_public_items<'a>(
...
+                crate_id: 0,
+                name: Some(path.name.clone()),
+                span: None,
+                visibility: rustdoc_types::Visibility::Public,
+                docs: None,
+                links: HashMap::new(),
+                attrs: Vec::new(),
+                deprecation: None,
+                inner: rustdoc_types::ItemEnum::Trait(rustdoc_types::Trait {
+                    is_auto: auto_traits.to_vec().contains(&path.name.as_str()),
```
obi1kenobi: Let's not do the `.to_vec().contains()` bit, and instead do this properly: extract a function for making trait items, have it take an `is_auto: bool` parameter, etc.

To make it easier to define these traits, I'd consider making a new private (file-level) struct like `ManualTraitItem` with only the subset of fields we need / care about, and then pass instances of that struct to the method that makes the "real" rustdoc item from that "description".

src/indexed_crate.rs
```diff
@@ -148,3 +156,73 @@ fn collect_public_items<'a>(
...
+                    is_unsafe: false,
+                    // The `item`, `generics`, `bounds` and `implementations`
+                    // are not present in the schema,
```
```suggestion
                    // are not currently present in the schema,
```

src/indexed_crate.rs
obi1kenobi: Assigning 0 is not safe. You can look up the correct ID in `crate_.paths` for each trait. I know that the `paths` field may be removed and replaced with something like `external_index`, but I sincerely hope that we're able to either eliminate this hack before that change happens or at least accommodate the change reasonably. Having an intentionally-incorrect crate ID seems like we're just setting ourselves up for much more trouble (and debugging) later.

Be careful, though: `core` and `std` are different crates with different crate IDs.

src/indexed_crate.rs
```diff
@@ -148,3 +156,73 @@ fn collect_public_items<'a>(
...
+                name: Some(path.name.clone()),
+                span: None,
+                visibility: rustdoc_types::Visibility::Public,
+                docs: None,
+                links: HashMap::new(),
+                attrs: Vec::new(),
+                deprecation: None,
+                inner: rustdoc_types::ItemEnum::Trait(rustdoc_types::Trait {
+                    is_auto: auto_traits.to_vec().contains(&path.name.as_str()),
+                    is_unsafe: false,
```
obi1kenobi: I think some of the traits are unsafe. `Send` and `Sync` come to mind.

src/adapter.rs
obi1kenobi: Yes, good catch.

src/adapter.rs
```diff
@@ -1163,7 +1163,21 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
...
+                                            // The needed foreign traits for lints are hand-written 
+                                            // in `dummy_trait_items`.
+                                            // See the comments in `src/indexed_crate.rs`
+                                            // for details.
+                                            let found_item = item_index
+                                                .get(&path.id)
+                                                .or_else(|| current_crate.dummy_trait_items.get(&path.id))
+                                                .or_else(|| previous_crate.and_then(|crate_| {
+                                                    crate_.dummy_trait_items.get(&path.id)
+                                                }));
```
obi1kenobi: ID collisions across rustdoc JSON can happen, so this is setting us up for some _really bad_ debugging time in the future.

Never do the "try your luck" ID matching. Always explicitly figure out whether the ID is in the baseline or the current crate, and then look up the appropriate one.

src/adapter.rs
obi1kenobi:
```suggestion
                                            // When the implemented trait is from the same crate
                                            // as its definition, the trait is expected to be present
                                            // in `item_index`. Otherwise, the
                                            // `rustdoc_types::Trait` is not in this rustdoc,
                                            // even if the trait is part of Rust `core` or `std`.
                                            // As a temporary workaround, some common
                                            // Rust built-in traits are manually "inlined"
                                            // with items stored in `manually_inlined_builtin_traits`.
```

src/adapter.rs
tonowak: Done in c25fe8efd190586972cbadb430e8840764be0de7.

src/indexed_crate.rs
tonowak: Done in 52c72aac72b9b642f944f07ce4bab9ddd7fb9bfb.

src/adapter.rs
tonowak: Done in 1f8530201a6544073f061823d91eb66d630914bf.

src/indexed_crate.rs
tonowak: Done in 0924a99007bedee1516fe6eae723072ba4d7ff5b.

src/adapter.rs
tonowak: Done in 14f01e0809c118d10b3c92a1acec463f686c1cdb.

tonowak: The comment changed, this conversation is no longer relevant.

src/indexed_crate.rs
tonowak: Fixed in 016398d8edbd2f7eb1dc87aebb0561ebb9ec957f.

src/indexed_crate.rs
tonowak: Done in 016398d8edbd2f7eb1dc87aebb0561ebb9ec957f.

src/indexed_crate.rs
tonowak: Done in 016398d8edbd2f7eb1dc87aebb0561ebb9ec957f.

src/indexed_crate.rs
tonowak: Done in 9add54e2554ce486e269549bb24cdd2f7234754b.

src/indexed_crate.rs
tonowak: Mentioned in https://github.com/obi1kenobi/cargo-semver-checks/issues/256.

tonowak:
This PR is ready for a new review iteration @obi1kenobi.

src/indexed_crate.rs
```diff
@@ -148,3 +166,142 @@ fn collect_public_items<'a>(
         Visibility::Crate | Visibility::Restricted { .. } => {}
     }
 }
+
+struct ManualTraitItem {
```
obi1kenobi, line 172:
```suggestion
#[derive(Debug)]
struct ManualTraitItem {
```
Essentially all types should be `Debug`, especially if it's as easy as adding a `derive`.

src/indexed_crate.rs
```diff
@@ -2,21 +2,39 @@ use std::collections::HashMap;
 
 use rustdoc_types::{Crate, Id, Item, Visibility};
 
+/// Besides the parsed rustdoc, it additionally contains
+/// some manually inlined `rustdoc_types::Trait`s of
+/// most common built-in traits.
+/// This is a temporary step, until we'll adopt rustdocs of foreign crates.
```
```suggestion
/// The rustdoc for a crate, together with associated indexed data to speed up common operations.
///
/// Besides the parsed rustdoc, it also contains some manually-inlined `rustdoc_types::Trait`s
/// of the most common built-in traits.
/// This is a temporary step, until we're able to combine rustdocs of multiple crates.
```
Small wording improvements + a headline that explains the type itself, since I felt it was a bit awkward to have a doc comment explaining a detail of the type without saying anything about the type as a whole.

obi1kenobi (commented):
Minor nitpicks, otherwise looks great.

obi1kenobi (approved):

PR #25, opened 2023-01-05
Add fields to Variant interface

staniewzki:
This PR adds fields edge to Variant interface, as described in https://github.com/obi1kenobi/cargo-semver-checks/issues/253.


obi1kenobi (approved):

obi1kenobi:
This is great. Would you mind making the matching PRs for the other rustdoc versions, including the new v24 one?

`git cherry-pick` of the commit onto the other branches should do the trick. It's possible that v24 needs a touch more code, since the representation of enums is slightly changed.

After we have merged all the PRs, we can release the new versions.

staniewzki:
Done!

PR #72, opened 2023-02-05
Add `explicit_reexport_of_matching_names` test case.

obi1kenobi:
Test case for rustdoc JSON underlying bug: https://github.com/rust-lang/rust/issues/107677


obi1kenobi:
The upstream issue was just fixed, so it should make it over to nightly very soon.

PR #172, opened 2023-05-03
Fix import resolution when local items overlap with glob imports

### LINKED ISSUE #171: False-positive: single module contains two items with the same name
This bug was originally found during the verification of semver-crater results, thanks to the goblin crate's [`SectionHeader` item](https://docs.rs/goblin/0.6.1/src/goblin/elf/section_header.rs.html#385).

It is most likely caused by faulty or missing implementation of shadowing rules when re-exporting items from their parent module.

A short demonstration of the problem can be found [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6397c410214a808229061148f12f08e1).

obi1kenobi:
Fixes #171.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        toolchain: ["1.68", "1.69", "1.70"]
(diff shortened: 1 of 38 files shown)


obi1kenobi:
Currently this is feature-complete, just needs quite a bit of cleanup and refactoring before merge.

PR #211, opened 2023-07-14
add AssociatedType to schema

u9g:


src/adapter/mod.rs
```diff
@@ -220,7 +223,7 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
         let coerce_to_type = coerce_to_type.clone();
         match type_name.as_ref() {
             "Item" | "Variant" | "FunctionLike" | "Importable" | "ImplOwner" | "RawType"
-            | "ResolvedPathType" | "GlobalValue" => {
+            | "ResolvedPathType" | "GlobalValue" | "AssociatedType" => {
```
obi1kenobi: I think this isn't necessary, `AssociatedType` isn't an interface so it shouldn't ever be in a situation where we're coercing _from_ it.
```suggestion
            | "ResolvedPathType" | "GlobalValue" => {
```

src/adapter/properties.rs
```diff
@@ -288,3 +289,22 @@ pub(crate) fn resolve_static_property<'a>(
...
+        "has_default" => resolve_property_with(
+            contexts,
+            field_property!(as_item, inner, {
+                let ItemEnum::AssocType {
+                    default, ..
+                } = &inner else { unreachable!("expected to have a AssocType") };
+                default.is_some().into()
+            }),
+        ),
+        _ => unreachable!("Static property {property_name}"),
```
obi1kenobi:
```suggestion
        _ => unreachable!("AssociatedType property {property_name}"),
```

src/rustdoc_schema.graphql
```diff
@@ -724,6 +726,14 @@ type PrimitiveType implements RawType {
   name: String!
 }
 
+"""
+https://docs.rs/rustdoc-types/latest/rustdoc_types/enum.ItemEnum.html#variant.AssocType
+"""
+type AssociatedType implements RawType {
```
```suggestion
type AssociatedType {
```
`RawType` is something else actually and it's going to get significantly reworked, it has some issues right now.

test_crates/traits_with_types/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#![feature(associated_type_defaults)]
```
obi1kenobi: Can you check if the rustdoc generation script works on a stable Rust release with this feature enabled?

If not, we might need to get creative or, worst case, just not have a default associated type test case if we can't fit it in.

test_crates/traits_with_types/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+publish = false
+name = "traits_with_types"
```
obi1kenobi: nit: `traits_with_types` is a bit ambiguous of a name, it doesn't immediately communicate what we're going for here.

How about naming the crate, its directory, and the test case file `traits_with_associated_types`?

src/indexed_crate.rs
```diff
@@ -1260,6 +1260,31 @@ mod tests {
             }
         }
 
+        #[test]
+        fn trait_types() {
```
obi1kenobi: Please keep the test case name matching the test crate name, so it's easier to correlate tests to the data they use.

src/indexed_crate.rs
```diff
@@ -1260,6 +1260,31 @@ mod tests {
...
+                })
+                .fold(std::collections::HashMap::new(), |mut acc, el| {
+                    let Some(v) = el.0 else { return acc };
+                    acc.insert(v.clone(), el.1.clone());
+                    acc
+                });
+
+            assert!(&associated_types["DeserializedType"].is_none());
+            assert!(&associated_types["SerializedType"].is_some());
+        }
```
obi1kenobi: This test is actually in the wrong submodule — it's in the `reexports` submodule even though it doesn't test re-exports at all.

We actually probably don't need this test at all, because it just checks something that's you already added a test for in `src/adapter/tests.rs`.

I'd just delete it.

src/rustdoc_schema.graphql
```diff
@@ -415,6 +415,8 @@ type Trait implements Item & Importable {
   method: [Method!]
 
   supertrait: [ImplementedTrait!]
+
+  associated_type: [AssociatedType!]!
```
```suggestion
  associated_type: [AssociatedType!]
```
because not all traits have associated types

src/rustdoc_schema.graphql
obi1kenobi: Also, this feels like it should be `implements Item` and therefore have all the item properties. Otherwise, we won't have access to spans, for example, which will make writing lints difficult.

obi1kenobi (commented):
A bunch of small nits that I think should be resolvable easily. Excited to get this merged soon 🚀

test_crates/traits_with_types/src/lib.rs
u9g: Switching rustup to stable then regenerating the rustdocs results in no errors, so I don't think this is a problem

src/rustdoc_schema.graphql
u9g: this is actually kind of confusing, the way `[AssociatedType!]!` reads to me is iterator of AssociatedType, the way `[AssociatedType!]` reads to me is iterator or null of AssociatedType, which doesn't really even make sense to me, would you be able to clarify how I should interpret this change?

src/adapter/mod.rs
u9g: fixed in 2329a50

src/adapter/properties.rs
u9g: fixed in a2aaaca

src/rustdoc_schema.graphql
u9g: Yeah, when implementing `RawType`, I don't know what was going through my head but definitely not common sense lol. I've removed that implementation, and we now implement Item in c792a1d

test_crates/traits_with_types/Cargo.toml
u9g: Sure, done in e01dadf

src/indexed_crate.rs
u9g: I just ended up removing this test in 58ce09b because of your next comment

src/indexed_crate.rs
u9g: Sure, done in 58ce09b

src/rustdoc_schema.graphql
u9g: I made this change regardless of that^ in a6e05ff

test_crates/traits_with_types/src/lib.rs
obi1kenobi: Nice! Yes, I see now that we're generating the rustdoc with `RUSTC_BOOTSTRAP` which is a bit sketchy but in this case the right thing to do 😅 Thanks for verifying!

src/rustdoc_schema.graphql
obi1kenobi: Here we're limited by the expressiveness of the GraphQL schema definition language (SDL), so we have to make do with what we've got.

There's no way in GraphQL SDL to say "iterator that possibly has zero elements," and unlike in GraphQL, there's never a need to worry about `null` on edges in Trustfall. So we make do — we appropriate the `null` value to mean "possibly no edges of this kind" as a matter of poorly-documented convention.

u9g: Oh okay, that makes perfect sense

PR #260, opened 2023-08-29
Implement properties around participation in public APIs on importable paths and items.

davidhewitt:
Related to https://github.com/obi1kenobi/cargo-semver-checks/issues/120

Adds new properties to `ImportablePath` to probe whether an API is "public", which is currently implemented as:
- Has an importable path
- Which may be deprecated
- If not deprecated, it must not be hidden

This was done together in collaboration with @obi1kenobi (thanks, was fun pairing!); one or both of us may pick this up again in the future time permitting. Mostly complete barring the TODOs in `edges.rs` and `indexed_crate.rs`, plus test output will change once those are corrected.

### DIFF
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -4,10 +4,7 @@ use trustfall::provider::{
     VertexIterator,
 };
 
-use crate::{
-    adapter::supported_item_kind, attributes::Attribute, indexed_crate::ImportablePath,
-    IndexedCrate,
-};
+use crate::{adapter::supported_item_kind, attributes::Attribute, IndexedCrate};
 
 use super::{optimizations, origin::Origin, vertex::Vertex, RustdocAdapter};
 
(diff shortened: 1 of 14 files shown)


obi1kenobi:
Summary of changes:
- Rebased to point to `rustdoc-v27` since that's currently the main branch of `trustfall-rustdoc-adapter`. My automation always backports from the latest version's branch to older versions.
- Propagated `#[doc(hidden)]` and deprecation info through the visibility tracking and indexing code. This resolves a few of the previously-outstanding TODOs marked in the code.
- Made the `ImportablePath` vertex variant hold an `Rc` for cheaper vertex cloning. This parallels the existing vertex variants' implementations.
- Added `doc_hidden`, `deprecated`, and `public_api_eligible` properties to the `Item` type. The first two have obvious implementations, the last one is similar to the one for `ImportablePath` but includes a visibility check (`pub` or default visibility, like enum variants) as well. This is necessary to properly handle cases like:
```rust
pub struct Example {
    #[doc(hidden)]
    pub not_actually_public_api: i64,
}
```

That case requires this new `public_api_eligible` field because struct fields are not importable themselves (one can't `use Example::not_actually_public_api`) which means they don't have access to a `public_api` property via `ImportablePath`. The way I intend to implement the check here is by saying "the contained field is eligible to be public API, and since its parent struct is public API, the field itself is public API as well."

PR #261, opened 2023-09-08
Add adapter support for a Module item type.

nmathewson:
Hello!

This is prerequisite for solving
https://github.com/obi1kenobi/cargo-semver-checks/issues/482

This is my first attempt at using trustfall, graphql, or the rustdoc_types crate, so I certainly may have missed some things.  Please do not assume I know what I am doing in the slightest. :wink: 

The patch is against rustdoc-v27.  I _think_ it should work similarly with the other branches since the Module struct is the same across all the supported versions of rustdoc, but I don't know how you prefer to handle porting patches like this.

### DIFF
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -162,6 +162,35 @@ pub(super) fn resolve_function_like_edge<'a>(
     }
 }
 
+pub(super) fn resolve_module_edge<'a>(
+    contexts: ContextIterator<'a, Vertex<'a>>,
+    edge_name: &str,
+    current_crate: &'a IndexedCrate<'a>,
+    previous_crate: Option<&'a IndexedCrate<'a>>,
+) -> ContextOutcomeIterator<'a, Vertex<'a>, VertexIterator<'a, Vertex<'a>>> {
+    match edge_name {
+        "item" => resolve_neighbors_with(contexts, move |vertex| {
+            let origin = vertex.origin;
+            let module_item = vertex.as_module().expect("vertex was not a Module");
+
+            let item_index = match origin {
+                Origin::CurrentCrate => &current_crate.inner.index,
+                Origin::PreviousCrate => {
+                    &previous_crate
+                        .expect("no previous crate provided")
+                        .inner
+                        .index
+                }
+            };
+
+            Box::new(module_item.items.iter().map(move |item_id| {
+                origin.make_item_vertex(item_index.get(item_id).expect("missing item"))
+            }))
+        }),
+        _ => unreachable!("resolve_module_edge {edge_name}"),
+    }
+}
+
 pub(super) fn resolve_struct_edge<'a>(
     contexts: ContextIterator<'a, Vertex<'a>>,
     edge_name: &str,
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -4,7 +4,7 @@ use trustfall::provider::{
     VertexIterator,
 };
 
-use crate::{attributes::Attribute, IndexedCrate};
+use crate::{adapter::supported_item_kind, attributes::Attribute, IndexedCrate};
 
 use super::{optimizations, origin::Origin, vertex::Vertex, RustdocAdapter};
 
@@ -186,6 +186,7 @@ pub(super) fn resolve_module_edge<'a>(
             Box::new(module_item.items.iter().filter_map(move |item_id| {
                 item_index
                     .get(item_id)
+                    .filter(|item| supported_item_kind(item))
                     .map(|item| origin.make_item_vertex(item))
             }))
         }),
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -183,8 +183,10 @@ pub(super) fn resolve_module_edge<'a>(
                 }
             };
 
-            Box::new(module_item.items.iter().map(move |item_id| {
-                origin.make_item_vertex(item_index.get(item_id).expect("missing item"))
+            Box::new(module_item.items.iter().filter_map(move |item_id| {
+                item_index
+                    .get(item_id)
+                    .map(|item| origin.make_item_vertex(item))
             }))
         }),
         _ => unreachable!("resolve_module_edge {edge_name}"),
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -35,6 +35,29 @@ pub(super) fn resolve_crate_edge<'a>(
 ) -> ContextOutcomeIterator<'a, Vertex<'a>, VertexIterator<'a, Vertex<'a>>> {
     match edge_name {
         "item" => optimizations::item_lookup::resolve_crate_items(adapter, contexts, resolve_info),
+        "root_module" => {
+            let current_crate = adapter.current_crate;
+            let previous_crate = adapter.previous_crate;
+
+            resolve_neighbors_with(contexts, move |vertex| {
+                let origin = vertex.origin;
+                let crate_ = vertex.as_crate().expect("vertex was not a crate!");
+                let item_index = match origin {
+                    Origin::CurrentCrate => &current_crate.inner.index,
+                    Origin::PreviousCrate => {
+                        &previous_crate
+                            .expect("no previous crate provided")
+                            .inner
+                            .index
+                    }
+                };
+
+                let module = item_index
+                    .get(&crate_.root)
+                    .expect("crate had no root module");
+                Box::new(std::iter::once(origin.make_item_vertex(module)))
+            })
+        }
         _ => unreachable!("resolve_crate_edge {edge_name}"),
     }
 }
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -66,7 +66,6 @@ pub(super) fn resolve_module_property<'a>(
     property_name: &str,
 ) -> ContextOutcomeIterator<'a, Vertex<'a>, FieldValue> {
     match property_name {
-        "is_crate" => resolve_property_with(contexts, field_property!(as_module, is_crate)),
         "is_stripped" => resolve_property_with(contexts, field_property!(as_module, is_stripped)),
         _ => unreachable!("Module property {property_name}"),
     }
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -61,6 +61,17 @@ pub(super) fn resolve_item_property<'a>(
     }
 }
 
+pub(super) fn resolve_module_property<'a>(
+    contexts: ContextIterator<'a, Vertex<'a>>,
+    property_name: &str,
+) -> ContextOutcomeIterator<'a, Vertex<'a>, FieldValue> {
+    match property_name {
+        "is_crate" => resolve_property_with(contexts, field_property!(as_module, is_crate)),
+        "is_stripped" => resolve_property_with(contexts, field_property!(as_module, is_stripped)),
+        _ => unreachable!("Module property {property_name}"),
+    }
+}
+
 pub(super) fn resolve_struct_property<'a>(
     contexts: ContextIterator<'a, Vertex<'a>>,
     property_name: &str,
--- a/src/adapter/tests.rs
+++ b/src/adapter/tests.rs
@@ -491,6 +491,39 @@ fn rustdoc_modules() {
         ],
         results
     );
+
+    let root_query = r#"
+{
+   Crate {
+       root_module {
+           module: name @output
+           is_crate @output
+           is_stripped @output
+           item @fold {
+               members: name @output
+               types: __typename @output
+           }
+       }
+   }
+}
+"#;
+
+    let results: Vec<Output> =
+        trustfall::execute_query(&schema, adapter.clone(), root_query, variables.clone())
+            .expect("failed to run query")
+            .map(|row| row.try_into_struct().expect("shape mismatch"))
+            .collect();
+
+    similar_asserts::assert_eq!(
+        vec![Output {
+            module: "modules".into(),
+            is_crate: true,
+            is_stripped: false,
+            members: vec![Some("hello".into()), Some("outer".into())],
+            types: vec!["Module".into(), "Module".into()],
+        }],
+        results
+    );
 }
 
 #[test]
--- a/src/adapter/tests.rs
+++ b/src/adapter/tests.rs
@@ -1,8 +1,8 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
 use anyhow::Context;
-use maplit::{btreemap, btreeset};
+use maplit::btreemap;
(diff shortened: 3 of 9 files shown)


src/rustdoc_schema.graphql
```diff
@@ -55,6 +55,36 @@ interface Item {
   span: Span
 }
 
+"""
+https://docs.rs/rustdoc-types/0.11.0/rustdoc_types/struct.Item.html
+https://docs.rs/rustdoc-types/0.16.0/rustdoc_types/struct.Module.html
```
nmathewson, line 65: I cited 0.16.0 here because 0.11.0 does not have the `is_stripped` field.

src/rustdoc_schema.graphql
```diff
@@ -55,6 +55,36 @@ interface Item {
...
+  # properties from Item
+  id: String!
+  crate_id: Int!
+  name: String
+  docs: String
+  attrs: [String!]!
+  visibility_limit: String!
+
+  # own properties
+  is_crate: Boolean!
```
nmathewson: It's possible that we don't even want to consider the top-level module (the one with is_crate=true) to be an item? But I think it's okay.

src/adapter/edges.rs
```diff
@@ -162,6 +162,35 @@ pub(super) fn resolve_function_like_edge<'a>(
...
+                    &previous_crate
+                        .expect("no previous crate provided")
+                        .inner
+                        .index
+                }
+            };
+
+            Box::new(module_item.items.iter().map(move |item_id| {
+                origin.make_item_vertex(item_index.get(item_id).expect("missing item"))
+            }))
```
obi1kenobi: I think we can trigger this `.expect()` with an item re-export from an external crate. Probably best to use `.filter_map()` to discard missing items from the iterator until we get a chance to properly handle cross-crate items at some point in the future.

src/adapter/properties.rs
```diff
@@ -61,6 +61,17 @@ pub(super) fn resolve_item_property<'a>(
     }
 }
 
+pub(super) fn resolve_module_property<'a>(
+    contexts: ContextIterator<'a, Vertex<'a>>,
+    property_name: &str,
+) -> ContextOutcomeIterator<'a, Vertex<'a>, FieldValue> {
+    match property_name {
+        "is_crate" => resolve_property_with(contexts, field_property!(as_module, is_crate)),
```
obi1kenobi: This is an entirely reasonable design here. But if you're up for it, I think we can take advantage of an advanced little Trustfall trick and do something clever and more ergonomic.

Here's the key observation that generally takes some getting used to when starting to use Trustfall: when modelling a data source in Trustfall, we don't actually have to map the underlying data representation one-to-one. We can model the _semantic_ representation of the underlying data, instead of matching how it is _syntactically_ represented in the underlying data source.

Here's what I mean.

Consider when we'd use the `is_crate` property: it's when we want to find the top-level module of the crate. The query for that might look like something like this:
```graphql
{
    Crate {
        item {
            ... on Module {
                is_crate @filter(op: "=", value: ["$true"])
                
                < ... etc ... >
            }
        }
    }
}
```

This makes sense, and is a one-to-one match for the data representation in the underlying JSON file. But it's also a bit clunky, no?

Consider this alternative (names subject to bikeshedding):
```graphql
{
    Crate {
        root_module {
            # this is already of type Module and represents the crate root
            < ... same stuff as the "etc." in the earlier query ... >
        }
    }
}
```

Now, `root_module` isn't an explicit relationship that is ever named as such in the underlying representation — everything in the JSON file goes through the item index — but this is what the _actual, truly-underlying data_ looks like, semantically. Nothing is stopping us from implementing such an edge in Trustfall, and it's often both a query ergonomics improvement _and_ it gives us some more decoupling from the specific implementation choices of the JSON representation.

I know that was a lot to take in at once. Did that make sense? Happy to answer any questions, or even hop on a quick video call to chat about it, if that would be helpful.

src/adapter/properties.rs
```diff
@@ -61,6 +61,17 @@ pub(super) fn resolve_item_property<'a>(
...
+        "is_stripped" => resolve_property_with(contexts, field_property!(as_module, is_stripped)),
```
obi1kenobi, line 69: `is_stripped` is another property of the JSON representation, and not of the underlying data model that JSON represents. More JSON elements have `is_stripped`, but I don't remember — do we expose it elsewhere?

It's fine to add, especially if we already expose it in other places. Just pointing it out as another example of the thing I mentioned in my comment above.

src/adapter/tests.rs
```diff
@@ -400,6 +400,116 @@ fn rustdoc_finds_statics() {
...
+
+    let mod_query = r#"
+{
+   Crate {
+       item {
+           ... on Module {
+               module: name @output
+               is_crate @output
+               is_stripped @output
+               item {
```
obi1kenobi: Adding `@fold` next to `item` might make this test a bit easier to read — it will aggregate the member item names into a list of strings for each module.

src/adapter/tests.rs
```diff
@@ -400,6 +400,116 @@ fn rustdoc_finds_statics() {
...
+        trustfall::execute_query(&schema, adapter.clone(), mod_query, variables.clone())
+            .expect("failed to run query")
+            .map(|row| row.try_into_struct().expect("shape mismatch"))
+            .collect();
+    results.sort_unstable();
+
+    similar_asserts::assert_eq!(
+        vec![
+            Output {
+                member: None,
```
obi1kenobi: Consider outputting `__typename` on the member items as well, to make it easier to verify that `None` member names aren't bugs and are instead item types that genuinely don't have their own names (like `impl` for example).

src/rustdoc_schema.graphql
obi1kenobi: I think we probably do want to consider it a module, since that both matches how Rust thinks about it and also because otherwise it'll be hard (perhaps even impossible) to answer "what items are at top level in the crate."

test_crates/modules/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+
```
obi1kenobi: minor nit: I'd just strip the leading and trailing whitespace in both of these test files, except for a single trailing newline at the end of all files

obi1kenobi (commented):
This is excellent work, truly A+ stuff 💯 I'm very impressed, because I know the sad state the Trustfall docs (and this project's docs) are in relative to the complexity involved here 😅 Thank you so much! 🙌

There's just one little entirely non-obvious thing missing, and that's to update `supported_item_kind()` in `adapter/mod.rs` to say that `ItemEnum::Module(..)` is a supported item kind. Without that, module items won't be indexed by our "look up items by name" optimization code. That in turn would lead to the strange behavior that enumerating all modules works fine but looking them up by their importable path incorrectly returns no results. This would be great to have a test case for, if you're up for it.

But again, there's zero way you could have spotted that, and the codebase didn't do anything whatsoever to direct you toward it. Sorry! And thank you for bearing with it 😁

Don't worry about porting this to earlier rustdoc versions. I have a script for that that does it mostly automagically, since the code is almost always identical or extremely close to it. (Yet another reason to bind to the semantics of the underlying data set instead of its bytes-on-disk representation.)

If you feel up for it, a subsequent PR adding docs around the sticking points you ran into would be super, super helpful and much appreciated. Also definitely good to mention that contributors shouldn't worry about porting their PRs to prior rustdoc versions and that I'm happy to sort that out.

src/adapter/properties.rs
nmathewson: I see `fields_stripped` on Struct and `variants_stripped` on Enum; I added this by analogy, though TBH I'm not 100% sure.

obi1kenobi:
Yup, it's totally fine to leave it then 👍

src/adapter/properties.rs
nmathewson: I think this is a fine idea, but I'm not 100% sure how to actually build this. I guess this would mean adding a "root_module" edge in the graphql file, and then adding a "root_module" handler in resolve_crate_items, but at that point I'm afraid I'm lost.  The best I can think to do, based on copy-and-paste, is something like:

```
	"root_module" =>
	    Box::new(
		optimizations::item_lookup::resolve_crate_items(adapter, contexts, resolve_info).filter(|item| {
                  // check if item is a module with is_crate == true
                })
	    ),
```

but I'm not really sure that's right.

src/adapter/tests.rs
nmathewson: I've tried to do this, but my trial-and-error didn't find a way to make this work well with the @fold I added. I would have liked to output `(member name, member type)` tuples, but I didn't know enough graphql.

src/adapter/properties.rs
obi1kenobi: That would work just fine, well done!

We can do slightly better by taking advantage of the JSON format structure: the JSON file has a key called `"root"` at top level whose value is the item ID of the root module. We can look up that ID in the crate index directly, which will let us avoid iterating through all items in the crate.

src/adapter/properties.rs
obi1kenobi: Yup, it's totally fine to leave it then 👍

src/adapter/tests.rs
obi1kenobi: Unfortunately we don't have a way for that yet, it's on my todo list but I haven't gotten to it.

`@fold` will do produce parallel lists of each property separately, so their types in `Output` would be `Vec<String>` for each. Trustfall also guarantees that the lists have equal length and are parallel, in the sense that the i-th element in each list corresponds to the same vertex each time.

nmathewson: Okay, I tried it that way, but I got: 

```
thread 'adapter::tests::rustdoc_modules' panicked at 'internal error: entered unreachable code: unexpected item.inner for item: Item { id: Id("0:9-0:11:1801"), crate_id: 0, name: None, span: Some(Span { filename: "src/outer.rs", begin: (4, 0), end: (4, 18) }), visibility: Public, docs: None, links: {}, attrs: [], deprecation: None, inner: Import(Import { source: "inner::T4", name: "T4", id: Some(Id("0:11:1801")), glob: false }) }', src/adapter/vertex.rs:62:22
```

As far as I can tell, I think this means that rustdoc_types::Import doesn't have a `__typename`, and so we can't examine it.  That's probably a bug, but not one I think I can fix in this branch.

src/adapter/edges.rs
```diff
@@ -162,6 +162,37 @@ pub(super) fn resolve_function_like_edge<'a>(
...
+                        .expect("no previous crate provided")
+                        .inner
+                        .index
+                }
+            };
+
+            Box::new(module_item.items.iter().filter_map(move |item_id| {
+                item_index
+                    .get(item_id)
+                    .map(|item| origin.make_item_vertex(item))
```
obi1kenobi: Throw a check for `supported_item_kind` here after resolving the item and before making the vertex, that's causing the error you're seeing in the test. Rustdoc's `Import` item type (which, confusingly, represents re-exports) isn't part of our schema at the moment so we shouldn't represent it as a vertex.

Now that I think of it, we should perhaps modify `make_item_vertex()` to call that function internally and return `Option<Vertex>` instead of `Vertex`... But that can go in a separate PR if you'd like, just to keep the scope here limited.

src/adapter/tests.rs
obi1kenobi: Ah that actually caught a different bug, see above.

src/adapter/properties.rs
nmathewson: Okay. I tried this:
```
	"root_module" => resolve_neighbors_with(contexts, move |vertex| {
	    let current_crate = adapter.current_crate;
	    let previous_crate = adapter.previous_crate;

	    let origin = vertex.origin;
	    let item_index = match origin {
                Origin::CurrentCrate => &current_crate.inner.index,
                Origin::PreviousCrate => {
		    &previous_crate
                        .expect("no previous crate provided")
                        .inner
                        .index
                }
	    };

	    let module = item_index.get(&Id("root".into())).expect("crate had no root module");
	    Box::new(std::iter::once(origin.make_item_vertex(module)))
	}),
```

but I'm afraid I ran into lifetime issues.  Do I need to pass current_crate and previous_crate  as arguments into this function manually?

obi1kenobi: This is a common issue: `adapter` itself doesn't necessarily live long enough to be moved in, but `adapter.current_crate` and `adapter.previous_crate` are `&'a` references so they live long enough. Assign them outside the closure and move them into it instead of moving `adapter` — almost all the other edge implementations elsewhere do the same thing for the same reason.

nmathewson: Ah, that did it!

nmathewson:
okay, I _think_ I did something for each of our open threads here; please let me know if there's more.  I'll also think about documentation a bit and circle around to it soon.

src/adapter/properties.rs
obi1kenobi: Given the new edge, I'd remove this property since I think it won't be that useful — if someone wants the crate-level module, they can get it directly via the new edge.

src/rustdoc_schema.graphql
```diff
@@ -27,6 +27,8 @@ type Crate {
   includes_private: Boolean!
   format_version: Int!
 
+  root_module: Module!
```
obi1kenobi, line 33: None of the other things have docs, which was just me being sloppy early on. But it's good to start going forward, and here it's easy enough.
```suggestion
  """
  The top-level module of the crate, in which everything else in the crate is contained.
  """
  root_module: Module!
```

obi1kenobi (commented):
Two minor nits and this is good to merge.

src/adapter/properties.rs
nmathewson: Hm. What if you wanted to get all of the modules _except_ for the root module? Would that justify this property?

obi1kenobi: That's still doable and in a way that more closely maps to the semantic data model (the query says "not the root module" explicitly here) instead of mapping to the `is_crate` field which is mostly for JSON representation purposes, not something Rust modules inherently have.

Here's how:
```graphql
{
    Crate {
        root_module {
            id @tag
        }

        item {
            ... on Module {
                id @filter(op: "!=", value: ["%id"])
                
                < ... these are all modules that aren't the root ... >
            }
        }
    }
}
```

obi1kenobi (approved):

obi1kenobi:
Nicely done! Thanks for putting this together!

obi1kenobi:
I've ported this PR to the other rustdoc versions, and queued up new patch releases for all supported rustdoc versions. When those patch releases make their way through the pipeline, you'll be able to use them in `cargo-semver-checks` by just running `cargo update` first.

It's possible it you might have to use `--precise` because some of our other dependencies raised MSRV to 1.70 and we haven't done that _yet_ so we can't use those newer versions _yet_.

Issue #279, opened 2023-10-06
Expand Constant to allow getting its Type


eugenesvk:
I'd like to get a list of all of crate's constants with their type and values, the latest changes allowed getting the values, but still not Type.

There is a workaround you suggested using Constant ID to find the required information, but it'd be nicer to be able to use the query language for that

You mentioned that this is pending Trustfall support for types

> Once Trustfall supports custom scalar types, it gives us the option of adding a "proper" type_ field to the schema that will be more ergonomic.

obi1kenobi:
Thanks for the issue and for the PR.

> as far as I understood, this is the ultimate crate that enables that

Thanks, it certainly tries to 😄 If you don't mind sharing, I'm curious to hear more! How did you come across this crate? What's the use case you are targeting?

eugenesvk:
**tldr**: I wanted to parse [Windows_sys ](https://docs.rs/windows-sys/latest/windows_sys/) crate for its **constants** (names/values), started with a cargo-semver-checks crate and after some digging through sparse docs and code realized that this adapter is the one providing all the schema descriptions


A bit **more context**: I wanted to have a more ergonomic way to use Windows **constants** to call Windows **APIs** in a **scripting** language:
  - `4097`  instead of some unreadable value that you have to lookup somewhere
  - `LOCALE_SENGLANGUAGE` you could just copy&paste the full name from the API MS documentation
  - `sEngLng` or even better use a more readable shorter name


Given that there are >100k constants, I couldn't load them dynamically within the language itself, so I thought I'd write a Rust library that stores all the values and returns them from a constant name. Found the Windows_sys crate that has all the contants, realized there is no reflection in Rust (so I can't just compile the crate and ask it to list all the contants), but someone at the Rust language forums suggested to look at cargo-semver-checks which parses cargo docs that contains that information.

So I've looked at the `pub_module_level_const_missing.ron` query, realized it's also missing :( the info, digged into what the semver checks is doing and ended up at this adapter crate

PR #280, opened 2023-10-06
Add Constant type/value properties

eugenesvk:
This is some draft of the edits I've made (currently only to v26 as that's the version of the locally generated docs) to be able to extract a list of constants from a crate as mentioned in https://github.com/obi1kenobi/trustfall-rustdoc-adapter/issues/279

Unresolved question: 
  - currently I'm using serde json to de/ser the [`Type` ](https://docs.rs/rustdoc-types/latest/rustdoc_types/enum.Type.html) since there is no conversion to [trustfall](https://docs.rs/trustfall/latest/trustfall/index.html)::[FieldValue](https://docs.rs/trustfall/latest/trustfall/enum.FieldValue.html#) implemented for this type, but that seems clunky, so what's the easiest way to have a proper support of this conversion (is it only possible in the trustfall main crate?)
    - it's clunky since on the receiving side I can only `try_into_struct`
```rs
  struct Output {
    // type_  	: rustdoc_types::Type, // Type→FieldValue not implemented, use ↓ json→string
    type_     	: String,
  }
let results:Vec<_> = trustfall::execute_query(&schema, adapter.clone(), query, variables.clone()).expect("failed to run query")
    .map(|row| row.try_into_struct::<Output>().expect("shape mismatch")).collect();
```
 into a string field, can't make it deserialise from json automatically).

(haven't done any formatting/tests yet)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ readme = "./README.md"
 [dependencies]
 trustfall = "0.6.0"
 rustdoc-types = "0.22.0"
+serde_json = "1.0.85"
 
 [dev-dependencies]
 anyhow = "1.0.58"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ readme = "./README.md"
 [dependencies]
 trustfall = "0.6.0"
 rustdoc-types = "0.22.0"
-serde_json = "1.0.85"
 
 [dev-dependencies]
 anyhow = "1.0.58"
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -465,6 +465,26 @@ pub(crate) fn resolve_associated_constant_property<'a>(
         _ => unreachable!("AssociatedConstant property {property_name}"),
     }
 }
+
+
+// const example from rustdocs
+// {"id"           	:"0:9:xx","crate_id":0,
+// "name"          	:"PUB_CONST_NAME",
+// "span"          	:{"filename":"","begin":[5,0],"end":[5,68]},
+// "visibility"    	:"public",
+// "docs"          	:null,
+// "links"         	:{},
+// "attrs"         	:[],
+// "deprecation"   	: null,
+// "inner"         	: { ← &inner will be set to ItemEnum::Constant(c), where c=struct rustdoc_types::Constant
+//   "constant"    	: { rustdoc_types::Constant has type_/expr/value/is_literal fields (docs.rs/rustdoc-types/latest/rustdoc_types/struct.Constant.html)
+//     "type"      	:{"primitive":"i32"}, or ↓ or any of docs.rs/rustdoc-types/latest/rustdoc_types/enum.Type.html
+//     "type"      	:{"borrowed_ref":{"lifetime": "'static","mutable": false,"type":{"primitive": "str"}}},
+//     "type"      	:{"resolved_path":{"name":"Years","id": "0:3:1633","args":{"angle_bracketed":{"args":[],"bindings":[]}}}},
+//     "expr"      	: "_",
+//     "value"     	: null,
+//     "is_literal"	: false}}
+// },
 pub(crate) fn resolve_constant_property<'a>(
     contexts: ContextIterator<'a, Vertex<'a>>,
     property_name: &str,
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -465,3 +465,24 @@ pub(crate) fn resolve_associated_constant_property<'a>(
         _ => unreachable!("AssociatedConstant property {property_name}"),
     }
 }
+pub(crate) fn resolve_constant_property<'a>(
+    contexts: ContextIterator<'a, Vertex<'a>>,
+    property_name: &str,
+) -> ContextOutcomeIterator<'a, Vertex<'a>, FieldValue> {
+    match property_name {
+        "type_"      => resolve_property_with(contexts, field_property!(as_item, inner, {
+            let ItemEnum::Constant(c) = &inner else {unreachable!("expected to have a Constant")};
+            // c.type_     .clone().into()}),), // Type→FieldValue not implemented, use ↓ json
+            serde_json::to_string(&c.type_.clone()).unwrap().into()}),),
+        "expr"       => resolve_property_with(contexts, field_property!(as_item, inner, {
+            let ItemEnum::Constant(c) = &inner else {unreachable!("expected to have a Constant")};
+            c.expr      .clone().into()}),),
+        "value"      => resolve_property_with(contexts, field_property!(as_item, inner, {
+            let ItemEnum::Constant(c) = &inner else {unreachable!("expected to have a Constant")};
+            c.value     .clone().into()}),),
+        "is_literal" => resolve_property_with(contexts, field_property!(as_item, inner, {
+            let ItemEnum::Constant(c) = &inner else {unreachable!("expected to have a Constant")};
+            c.is_literal.clone().into()}),),
+        _ => unreachable!("Constant property {property_name}"),
+    }
+}
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -502,7 +502,7 @@ pub(crate) fn resolve_constant_property<'a>(
             c.value     .clone().into()}),),
         "is_literal" => resolve_property_with(contexts, field_property!(as_item, inner, {
             let ItemEnum::Constant(c) = &inner else {unreachable!("expected to have a Constant")};
-            c.is_literal.clone().into()}),),
+            c.is_literal        .into()}),),
         _ => unreachable!("Constant property {property_name}"),
     }
 }
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -462,16 +462,6 @@ pub(crate) fn resolve_associated_constant_property<'a>(
                 default.clone().into()
             }),
         ),
-        "type_" => resolve_property_with(
-            contexts,
-            field_property!(as_item, inner, {
-                let ItemEnum::AssocConst { type_, .. } = &inner else {
-                    unreachable!("expected to have a AssocConst")
-                };
-                // type_.clone().into() // Type→FieldValue not implemented, use ↓ json
-                serde_json::to_string(&type_.clone()).unwrap().into()
-            }),
-        ),
         _ => unreachable!("AssociatedConstant property {property_name}"),
     }
 }
@@ -487,6 +477,7 @@ pub(crate) fn resolve_associated_constant_property<'a>(
 // "deprecation"   	: null,
 // "inner"         	: { ← &inner will be set to ItemEnum::Constant(c), where c=struct rustdoc_types::Constant
 //   "constant"    	: { rustdoc_types::Constant has type_/expr/value/is_literal fields (docs.rs/rustdoc-types/latest/rustdoc_types/struct.Constant.html)
+//     "type"      	: currently not supported as a property pending custom scalar type support in Trustfall (https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/280#discussion_r1348950734)
 //     "type"      	:{"primitive":"i32"}, or ↓ or any of docs.rs/rustdoc-types/latest/rustdoc_types/enum.Type.html
 //     "type"      	:{"borrowed_ref":{"lifetime": "'static","mutable": false,"type":{"primitive": "str"}}},
 //     "type"      	:{"resolved_path":{"name":"Years","id": "0:3:1633","args":{"angle_bracketed":{"args":[],"bindings":[]}}}},
@@ -499,16 +490,6 @@ pub(crate) fn resolve_constant_property<'a>(
     property_name: &str,
 ) -> ContextOutcomeIterator<'a, Vertex<'a>, FieldValue> {
     match property_name {
-        "type_" => resolve_property_with(
-            contexts,
-            field_property!(as_item, inner, {
-                let ItemEnum::Constant(c) = &inner else {
-                    unreachable!("expected to have a Constant")
-                };
-                // c.type_     .clone().into()}),), // Type→FieldValue not implemented, use ↓ json
-                serde_json::to_string(&c.type_.clone()).unwrap().into()
-            }),
-        ),
         "expr" => resolve_property_with(
             contexts,
             field_property!(as_item, inner, {
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -462,6 +462,16 @@ pub(crate) fn resolve_associated_constant_property<'a>(
                 default.clone().into()
             }),
         ),
+        "type_" => resolve_property_with(
+            contexts,
+            field_property!(as_item, inner, {
+                let ItemEnum::AssocConst { type_, .. } = &inner else {
+                    unreachable!("expected to have a AssocConst")
+                };
+                // type_.clone().into() // Type→FieldValue not implemented, use ↓ json
+                serde_json::to_string(&type_.clone()).unwrap().into()
+            }),
+        ),
         _ => unreachable!("AssociatedConstant property {property_name}"),
     }
 }
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -466,25 +466,6 @@ pub(crate) fn resolve_associated_constant_property<'a>(
     }
 }
 
-// const example from rustdocs
-// {"id"           	:"0:9:xx","crate_id":0,
-// "name"          	:"PUB_CONST_NAME",
-// "span"          	:{"filename":"","begin":[5,0],"end":[5,68]},
-// "visibility"    	:"public",
-// "docs"          	:null,
-// "links"         	:{},
-// "attrs"         	:[],
-// "deprecation"   	: null,
-// "inner"         	: { ← &inner will be set to ItemEnum::Constant(c), where c=struct rustdoc_types::Constant
-//   "constant"    	: { rustdoc_types::Constant has type_/expr/value/is_literal fields (docs.rs/rustdoc-types/latest/rustdoc_types/struct.Constant.html)
-//     "type"      	: currently not supported as a property pending custom scalar type support in Trustfall (https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/280#discussion_r1348950734)
-//     "type"      	:{"primitive":"i32"}, or ↓ or any of docs.rs/rustdoc-types/latest/rustdoc_types/enum.Type.html
-//     "type"      	:{"borrowed_ref":{"lifetime": "'static","mutable": false,"type":{"primitive": "str"}}},
-//     "type"      	:{"resolved_path":{"name":"Years","id": "0:3:1633","args":{"angle_bracketed":{"args":[],"bindings":[]}}}},
-//     "expr"      	: "_",
-//     "value"     	: null,
-//     "is_literal"	: false}}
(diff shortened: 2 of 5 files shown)


src/rustdoc_schema.graphql
```diff
@@ -694,6 +694,37 @@ type Constant implements Item & Importable & GlobalValue {
...
+  A type represented in the "raw" rustdoc JSON representation, see interface RawType.
+  Currently serialized as a serde_json String instead of a proper RawType
+  ```rust
+  //          	                	            	// rustdocs type field
+  const MIN   	:          usize	= 16       ;	// {"primitive":"usize"}
+  const LOG_AS	: &'static str  	= "batch"  ;	// {"borrowed_ref":{"lifetime":"'static","mutable":false,"type":{"primitive":"str"}}}
+  const YEAR  	:          Years	= Years(42);	// {"resolved_path":{"name":"Years","id":"0:3:1633","args":{"angle_bracketed":{"args":[],"bindings":[]}}}}
+  ```
+  """
+  type_: String
```
obi1kenobi: I'd advise against this, and in favor of an alternative approach I think is more elegant.

Instead of needing code to unpack this kind of serde-in-serde representation, consider having your use case grab the `id` value here and use it to look up the item in the [`index` of the deserialized rustdoc value](https://docs.rs/rustdoc-types/latest/rustdoc_types/struct.Crate.html#structfield.index) — the same one backing the queries here.

That way has a few advantages:
- You'll get a type-safe way to interact with the complex type information of constants.
- The code won't need to sneak serialized JSON into a string field, which as you've noted causes ergonomics problems.
- Once Trustfall supports custom scalar types, it gives us the option of adding a "proper" `type_` field to the schema that will be more ergonomic. (No concrete timeline on this, though I'd love to direct my time toward building it if some funding were available.)

src/rustdoc_schema.graphql
```diff
@@ -694,6 +694,37 @@ type Constant implements Item & Importable & GlobalValue {
...
+  const LOG_AS  	: &'static str  	= "batch"  ;	// true      	None       	`"\"batch\""` including escaped quotes
+  const YEAR    	:          Years	= Years(42);	// false     	None       	"_"
+  const EXPR_2_2	:          i32  	= 2 + 2    ;	// false     	`"4i32"    	"_"
+  const FN_FIVE 	:          i32  	= five()   ;	// false     	`"5i32"    	"_"
+  const fn five() -> i32 { 5 } ;
+  struct Years(i32);
+  ```
+  If the constant is set to be equal to another constant, the expr holds the name of that other constant, while the value holds the value of that other constant.
+  If the constant is set by evaluating a `const` expression, such as `2 + 2` or a `const fn` call, rustdoc's current behavior is to show an expr of `"_"` (the value is evaluated)
+  """
```
obi1kenobi: This comment is excellent! Thanks for looking into and documenting all the edge cases.

As it's currently positioned, it technically only applies to the `expr` property here so editor integrations won't show it for `value` or `is_literal` even though it's clearly applicable there. Any thoughts on how to tweak it to document the other properties as well?

I can see a couple of options:
- Move the docs to apply to the `Constant` type itself, then add docs on each property with a brief explanation + a mention that the type docs have a complete example covering all the edge cases.
- Mostly duplicate that doc comment to each property, emphasizing the specific property's values in each doc.

src/rustdoc_schema.graphql
```diff
@@ -694,6 +694,37 @@ type Constant implements Item & Importable & GlobalValue {
...
+  The expression/value of the constant, if any, as a Rust literal, expression, or `"_"`, as well as its literal flag. For example:
+  ```rust
+  //            	                	            	// is_literal	value      	expr
+  const MIN     	:          usize	= 16       ;	// true      	`"16usize"`	`16`
+  const MIN_SIZE	:          usize	= MIN      ;	// false     	`"16usize"`	"MIN", referring to the other constant's name
+  const LOG_AS  	: &'static str  	= "batch"  ;	// true      	None       	`"\"batch\""` including escaped quotes
+  const YEAR    	:          Years	= Years(42);	// false     	None       	"_"
+  const EXPR_2_2	:          i32  	= 2 + 2    ;	// false     	`"4i32"    	"_"
+  const FN_FIVE 	:          i32  	= five()   ;	// false     	`"5i32"    	"_"
```
obi1kenobi: Let's try to make this representation a bit narrower, for example by reducing the spacing in the Rust code. Docs are often displayed in a side panel or on hover, where the panel can often be well under 100 chars. We don't have to squeeze every last char of width out (and the alignment here is very nice!) but we should avoid spending 10 spaces where 1-2 would do.

obi1kenobi (commented):
Nicely done! Thanks for the PR, I'd be happy to merge something like this.

I have a few suggestions around how to get the type info out more conveniently, and around some of the docs structure. Nothing major, should be resolvable in just a couple of minutes.

Also, are you interested in associated constants as well, or just regular ones? If the associated constants are missing fields, I'd be happy to take a PR there as well.

src/rustdoc_schema.graphql
```diff
@@ -694,6 +694,37 @@ type Constant implements Item & Importable & GlobalValue {
   docs: String
   attrs: [String!]!
   visibility_limit: String!
+  # properties for Constant
```
obi1kenobi, line 698:
```suggestion

  # properties for Constant
```

obi1kenobi: Just adding a blank line between the groups of properties.

src/rustdoc_schema.graphql
eugenesvk: should be better now

By the way, where do these docs go? Checked cargo docs and the crate docs site and don't see any schema docs

(oh, but then remembered you have docs in the online playground, so I guess that's where they are?)

src/rustdoc_schema.graphql
eugenesvk: Hard for me to decide without being able to see how the docs appear "in practice", how do I check that locally?

I've added a brief note that refers to the `expr`, but now that you mention editor integrations that might not be good enough as you'd need a per-property doc?

If so, then then your second suggestion is better, but then I don't think it's much of help to have a table of each 3 properties in the general doc, so this should also be split into only including the relevant value

If not and if the docs are laid out sequentialy like on docs.rs, then I think it's easier to have a single "comprehensive example" with a brief not telling the reader to look up

eugenesvk:
> Also, are you interested in associated constants as well, or just regular ones? If the associated constants are missing fields, I'd be happy to take a PR there as wel

just the regular ones, the associated ones already have the default value though seems to miss the `type_` field, so added that as well

AssocConst
Fields
type_: [Type](https://docs.rs/rustdoc-types/latest/rustdoc_types/enum.Type.html)
default: [Option](https://doc.rust-lang.org/nightly/core/option/enum.Option.html)<[String](https://doc.rust-lang.org/nightly/alloc/string/struct.String.html)>

src/rustdoc_schema.graphql
obi1kenobi: They go in the "schema" pane of playgrounds — in this case, https://play.predr.ag/rustdoc which unfortunately at the moment has broken side scrolling because I suck at front-end engineering 😅 

image

They are also shown on hover in editors with a language server for GraphQL, just like rust-analyzer shows docs on hover for Rust code.

src/rustdoc_schema.graphql
```diff
@@ -878,14 +916,36 @@ type AssociatedConstant implements Item {
...
   For example:
   ```rust
-  trait BatchIterator<const MIN: usize> {
-    const SIZE: usize = 16;  // `"16"` is the default
-    const LOG_AS: &'static str = "batch";  // `"\"batch\""` is the default, including escaped quotes
-    const MIN_SIZE: usize = MIN;  // "MIN" is the default, referring to the other constant's name
+  const fn five() -> i32 { 5 };
+  struct Years(i32);
+  trait BatchIterator<const MIN: usize> {	    // rustdocs default field
+    const MIN     : usize        = 16       ; // `"16"`
```
obi1kenobi: The const generic and the associated const have conflicting names here.

src/rustdoc_schema.graphql
```diff
@@ -878,14 +916,36 @@ type AssociatedConstant implements Item {
...
+    const MIN_SIZE: usize        = MIN      ; // `"MIN"`, referring to the other constant's name
+    const LOG_AS  : &'static str = "batch"  ; // `"\"batch\""`, including escaped quotes
+    const EXPR2_2 : i32          = 2+2      ; // `"_"`
+    const FN_FIVE : i32          = five()   ; // `"_"`
+    const YEAR    : Years        = Years(42); // `"_"`
```
obi1kenobi: The added constants don't make a ton of sense in the context of `BatchIterator` as a trait name. In such a case, perhaps it's best to not try to make any of the names make sense and just change the trait name to `Example` etc.

src/rustdoc_schema.graphql
```diff
@@ -878,14 +916,36 @@ type AssociatedConstant implements Item {
...
+    const MIN     : usize        = 16       ; // {"primitive":"usize"}
+    const MIN_SIZE: usize        = MIN      ; // {"primitive":"usize"}
+    const LOG_AS  : &'static str = "batch"  ; // {"borrowed_ref":{"lifetime":null,"mutable":false,"type":{"primitive":"str"}}}
+    const YEAR    : Years        = Years(42); // {"resolved_path":{"name":"Years","id":"0:3:1633","args":{"angle_bracketed":{"args":[],"bindings":[]}}}}
+    const EXPR2_2 : i32          = 2+2      ; // {"primitive":"i32"}
+    const FN_FIVE : i32          = five()   ; // {"primitive":"i32"}
+  }
+  ```
+  """
+  type_: String
```
obi1kenobi: I don't think adding `type_` as a property in either const or associated const is a good idea. It isn't in line with the rest of the crate's approach, and it prevents us from doing something better later.

I'd prefer to not add it. This shouldn't limit your ability to extract that information via the ID + index lookup method I mentioned.

src/rustdoc_schema.graphql
eugenesvk: Ok, so I've set up a local editor LSP and can see the comments live

But then within Rust codeblocks the single backticks markdown formatting is ignored (makes sense since this is regular rust comments now, and even doc comments aren't converted since this isn't done via cargo but LSP)
```
type Item;  // `"Item"` is the name of the associated type
```

so I'll remove them as they serve no purpose here?

And re. your previous ?, since there is no guarantee that all fields are present in an actual query, will go with your 2nd dupe option

src/rustdoc_schema.graphql
```diff
@@ -695,6 +695,44 @@ type Constant implements Item & Importable & GlobalValue {
...
+  """
+  expr: String
+  """
+  See `expr` docs for a comprehensive example also covering `value`
+  """
+  value: String
+  """
+  See `expr` docs for a comprehensive example also covering `is_literal`
+  """
+  is_literal: Boolean
```
obi1kenobi: Any particular reason why you decided to go with putting the docs in the `expr` property instead of one of the other options? To me it seems less helpful than the other options, but perhaps I've missed something.

src/rustdoc_schema.graphql
eugenesvk: changed the trait name

src/rustdoc_schema.graphql
eugenesvk: renamed

src/rustdoc_schema.graphql
eugenesvk: ok, will check the id way, meanwhile removed type from Assoc/Const properties

src/rustdoc_schema.graphql
eugenesvk: Yeah, that was before I've reviewed your ID suggestion, removed types

src/rustdoc_schema.graphql
```diff
@@ -695,6 +695,61 @@ type Constant implements Item & Importable & GlobalValue {
...
+  const MIN_SIZE: usize = MIN      ; // "MIN", referring to the other constant's name
+  const LOG_AS  : &str  = "batch"  ; // "\"batch\"", including escaped quotes
+  const YEAR    : Years = Years(42); // "_"
+  const EXPR_2_2: i32   = 2 + 2    ; // "_"
+  const FN_FIVE : i32   = five()   ; // "_"
+  const fn five() -> i32 { 5 };
+  struct Years(i32);
+  ```
+  If the constant is set:
+
```
obi1kenobi, line 713: minor formatting nit:
```suggestion

  If the constant is set:
```

src/rustdoc_schema.graphql
```diff
@@ -695,6 +695,61 @@ type Constant implements Item & Importable & GlobalValue {
...
+  const MIN_SIZE: usize = MIN      ; // "16usize"
+  const LOG_AS  : &str  = "batch"  ; // None
+  const YEAR    : Years = Years(42); // None
+  const EXPR_2_2: i32   = 2 + 2    ; // "4i32"
+  const FN_FIVE : i32   = five()   ; // "5i32"
+  const fn five() -> i32 { 5 };
+  struct Years(i32);
+  ```
+  If the constant is set:
+
```
obi1kenobi, line 732:
```suggestion

  If the constant is set:
```

src/rustdoc_schema.graphql
```diff
@@ -882,21 +937,24 @@ type AssociatedConstant implements Item {
 
   For example:
   ```rust
-  trait BatchIterator<const MIN: usize> {
-    const SIZE: usize = 16;  // `"16"` is the default
-    const LOG_AS: &'static str = "batch";  // `"\"batch\""` is the default, including escaped quotes
-    const MIN_SIZE: usize = MIN;  // "MIN" is the default, referring to the other constant's name
```
obi1kenobi, line 888: In the original example, `MIN_SIZE` was referring to the const generic constant. In the new example it refers to a local sibling constant only, and there's no example with a const generic. Mind adding a const generic example as well?

src/rustdoc_schema.graphql
```diff
@@ -882,21 +937,24 @@ type AssociatedConstant implements Item {
...
+    const YEAR    : Years        = Years(42); // `"_"`
   }
   ```
 
-  If the associated constant is on a type's inherent impl, the default is always required to be set.
+  If the associated constant is on a type's inherent impl, `default` is always required to be set.
 
-  If the associated constant is set to be equal to another constant, the default holds the name
-  of that other constant.
+  If the associated constant is set
```
```suggestion
  If the associated constant is set:
```

src/rustdoc_schema.graphql
```diff
@@ -882,21 +937,24 @@ type AssociatedConstant implements Item {
...
 
-  If the associated constant is set by evaluating a `const` expression, such as `2 + 2` or
-  a `const fn` call, rustdoc's current behavior is to show a default value of `"_"`
-  instead of evaluating the constant value or including the full expression.
+    - to be equal to another constant, `default` holds the name of that other constant.
+    - by evaluating a `const` expression, such as `2 + 2` or a `const fn` call, `default` is `"_"` instead of evaluating the constant value or including the full expression.
```
obi1kenobi: just for line length
```suggestion
    - by evaluating a `const` expression, such as `2 + 2` or a `const fn` call, 
      `default` is `"_"` instead of evaluating the constant value or including the full expression.
```

Cargo.toml
```diff
@@ -13,6 +13,7 @@ readme = "./README.md"
 [dependencies]
 trustfall = "0.6.0"
 rustdoc-types = "0.22.0"
+serde_json = "1.0.85"
```
obi1kenobi:
```suggestion
```
I don't think we need `serde_json` as a dependency anymore.

src/adapter/properties.rs
```diff
@@ -465,3 +465,58 @@ pub(crate) fn resolve_associated_constant_property<'a>(
...
+// "inner"         	: { ← &inner will be set to ItemEnum::Constant(c), where c=struct rustdoc_types::Constant
+//   "constant"    	: { rustdoc_types::Constant has type_/expr/value/is_literal fields (docs.rs/rustdoc-types/latest/rustdoc_types/struct.Constant.html)
+//     "type"      	: currently not supported as a property pending custom scalar type support in Trustfall (https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/280#discussion_r1348950734)
+//     "type"      	:{"primitive":"i32"}, or ↓ or any of docs.rs/rustdoc-types/latest/rustdoc_types/enum.Type.html
+//     "type"      	:{"borrowed_ref":{"lifetime": "'static","mutable": false,"type":{"primitive": "str"}}},
+//     "type"      	:{"resolved_path":{"name":"Years","id": "0:3:1633","args":{"angle_bracketed":{"args":[],"bindings":[]}}}},
+//     "expr"      	: "_",
+//     "value"     	: null,
+//     "is_literal"	: false}}
+// },
```
obi1kenobi: Would you mind if we removed this block? While it's nice to see what the underlying representation is, the JSON format is quite unstable (it changes on average once per Rust version) and I have a strong feeling that we'll forget to keep this updated to match the current format. The only thing worse than no examples is wrong examples, and I think this will very quickly become a wrong example.

Also, even on day 1 it will require lots of extra manual work to recreate this example in all the other `trustfall-rustdoc-adapter` versions I maintain, and I won't be able to use my usual PR-porting script to do it.

```suggestion
```

obi1kenobi (commented):
Just a couple of nitpicks on docs and such, otherwise it's good to merge. I usually don't merge PRs marked "draft" so when you've had a chance to take one more look and feel it's good to merge on your end, move it out of the draft state and I'll give it another quick look as well and then merge + publish.

src/rustdoc_schema.graphql
eugenesvk: This is actually intentional since no space after the "header" breaks the formatting, so instead of 2 bullet points you get one line (at least in my Sublime editor with LSP plugin)

Not sure how it works in your wasm playground or any other tools

Let me know if you want to add an extra empty line before anyway (personally prefer a bit more compact, but up to you)

src/rustdoc_schema.graphql
eugenesvk: restored MIN to be the const generic and the sibling is just a standalone NUM number

eugenesvk:
Done the last thing I wanted -  updated a test, though had to split logic for Constants with more properties like `expr`/`value` and for GlobalValues that don't have those and only test name/path as before

(although I've only run the local test for this updaed const crate, not the whole test suite)

Then pending the unresolved doc formatting edits should be more or less fine?

And as far as I understood, porting to other rustdocs versions is semi-auto-done with the help of your magic script?

src/rustdoc_schema.graphql
obi1kenobi: Ah nice, thanks for double-checking it in the editor.

Let's add the extra empty line before the block, though. I think that way it's a bit more readable when it's not rendered as markdown.

obi1kenobi: It's just the adding the newline, so I'll merge this as-is and add the newline in a separate PR and then cut a release so we can move faster.

obi1kenobi:
Looks great. Going to merge this, add the newline in the docs, and cut a new release that you should be able to use.

Thanks for the help! Let me know how this all works out in your use case.

> And as far as I understood, porting to other rustdocs versions is semi-auto-done with the help of your magic script?

Yup! I'll take care of porting it and cut a release for all the rustdoc versions that `cargo-semver-checks` currently supports: v24, v26, v27.

obi1kenobi:
New versions are available on crates.io!

PR #299, opened 2023-11-07
Implement properties around participation in public APIs on importable paths and items. (#260)

obi1kenobi:
* implement notion of 'public_api' for importable paths

* Propagate deprecation and doc(hidden) info through visibility tracking.

* Rc the ImportablePath inside the vertex enum.

* Add public API eligibility on items for nested items' API visibility.

* Add test coverage for enum variants and trait associated items.

---------

Co-authored-by: Predrag Gruevski <obi1kenobi82@gmail.com>


obi1kenobi:
Unfortunately, rustdoc JSON had a blocking bug prior to Rust 1.73 that will prevent this from working correctly.

We'll have to drop rustdoc v24 support to ship this feature.

PR #307, opened 2023-12-29
`implemented_trait` was implemented

SuperSonicHub1:
Noticed that a TODO wasn't removed from the docs while writing a query in the playground.


obi1kenobi (approved):
Neat! I forgot we implemented this, but [apparently we did](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/rustdoc-v28/src/adapter/edges.rs#L354-L398).

Thanks for digging in and fixing this! If you're interested in digging in deeper (no pressure), I'd be happy to suggest things to look into.

PR #309, opened 2024-01-10
add union types to schema

### LINKED ISSUE #36: Implement the `Union` type, making union types queriable
They are `Item`, `Importable`, and `ImplOwner`. They also have fields, probably best represented as type `StructField`.

fprasx:
- [x] Add `union` to schema
- [x] Update adapter

Fixes #36.

### DIFF
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -284,6 +284,7 @@ pub(crate) fn supported_item_kind(item: &Item) -> bool {
             | rustdoc_types::ItemEnum::StructField(..)
             | rustdoc_types::ItemEnum::Enum(..)
             | rustdoc_types::ItemEnum::Variant(..)
+            | rustdoc_types::ItemEnum::Union(..)
             | rustdoc_types::ItemEnum::Function(..)
             | rustdoc_types::ItemEnum::Impl(..)
             | rustdoc_types::ItemEnum::Trait(..)
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -92,8 +92,8 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                 "Crate" => properties::resolve_crate_property(contexts, property_name),
                 "Item" => properties::resolve_item_property(contexts, property_name),
                 "ImplOwner" | "Struct" | "StructField" | "Enum" | "Variant" | "PlainVariant"
-                | "TupleVariant" | "StructVariant" | "Trait" | "Function" | "Method" | "Impl"
-                | "GlobalValue" | "Constant" | "Static" | "AssociatedType"
+                | "TupleVariant" | "StructVariant" | "Union" | "Trait" | "Function" | "Method"
+                | "Impl" | "GlobalValue" | "Constant" | "Static" | "AssociatedType"
                 | "AssociatedConstant" | "Module"
                     if matches!(
                         property_name.as_ref(),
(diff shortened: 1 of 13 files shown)


obi1kenobi:
The schema is reasonable :+1:

Make sure to both update the adapter and the index lookup optimizations, otherwise we might end up in a situation where index lookups fail to find unions that otherwise get returned when the optimization isn't used.

Lmk if you get stuck on anything or would like to chat about anything.

fprasx:
Sounds good. I'm basically just grepping for `union/struct` and adding union.

fprasx:
Ok, the meat of it is there. 

I'm not too familiar with the codebase, do you know if there are any specific union-related tests I should add, off the top of your head?

src/adapter/mod.rs
```diff
@@ -254,7 +261,7 @@ impl<'a> Adapter<'a> for RustdocAdapter<'a> {
                             actual_type_name,
                             "PlainVariant" | "TupleVariant" | "StructVariant"
                         ),
-                        "ImplOwner" => matches!(actual_type_name, "Struct" | "Enum"),
+                        "ImplOwner" => matches!(actual_type_name, "Struct" | "Enum" | "Union"),
```
obi1kenobi, line 264: (GitHub doesn't allow commenting on lines not modified in the PR)

Make sure to also include `ItemEnum::Union` in `supported_item_kind` on line 280 in this file. Otherwise you'll get the exact "unions looked up by name don't exist" bug I mentioned.

obi1kenobi (commented):
> do you know if there are any specific union-related tests I should add

I'd add a test crate with some unions, and then make sure:
- querying for all unions and their fields returns correct data
- querying for union importable paths returns correct data (even with re-exports)
- querying unions by importable path returns correct data (this will use the fast paths and use an index lookup optimization)

src/adapter/mod.rs
fprasx: Shoot, I was grepping for '\"Enum\"' so I missed this. Fixed now.

obi1kenobi: It happens! You would have caught it in a test though :) Also, this is why we do code review!

fprasx:
Hmm, it actually seems less disruptive to just test `union` where `struct/enum` are getting tested in the crates that exist right now, for example:
> querying for union importable paths returns correct data (even with re-exports)

can be tested in the `importable_paths` test crate, where it's tested for `struct/enum`

obi1kenobi:
Let me know when this is ready for another review!

obi1kenobi:
> can be tested in the `importable_paths` test crate, where it's tested for `struct/enum`

btw, make sure to check out that test case, because I'm not sure it's testing quite what we might be assuming it's testing here

fprasx:
You're right, I did some more digging into the test crates, and that one is testing something else. I've actually come to your original conclusion post-dig, which is that it's just simpler to have a crate for unions. Because structs and enums were there from the beginning, there are no specific tests for them, rather they're just used everywhere. Since unions are new their introduction might cause some bugs so they can get \~their own crate\~.

I've gotta a pretty busy day today, so I might not be able to finish this up today, but I'll try to get it done by the weekend!

obi1kenobi:
Sounds good! Thanks for setting expectations explicitly as well, it'll help me plan my own day as well.

fprasx:
Ok, added a test crate. I think I've covered everything (importing, visibility, field info), but not sure.

obi1kenobi (approved):
Awesome, ship it! 🚀

obi1kenobi:
I've used a bit of magic automation to port this PR to `rustdoc-v27` and `rustdoc-v26`, and I'll release new versions for all three so that the next `cargo update` in `cargo-semver-checks` picks them up automatically.

PR #310, opened 2024-01-10
Add `rust-toolchain.toml` file to ensure generated rustdoc has correct version.

fprasx:
I was getting errors like the following:

```
---- adapter::tests::rustdoc_associated_consts stdout ---- 
thread 'adapter::tests::rustdoc_associated_consts' panicked at src/adapter/tests.rs:580:49:
failed to parse rustdoc: Error("missing field `is_object_safe`", line: 1, column: 99832)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- adapter::tests::importable_paths stdout ----
thread 'adapter::tests::importable_paths' panicked at src/adapter/tests.rs:836:49:
failed to parse rustdoc: Error("missing field `is_object_safe`", line: 1, column: 165267)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

because I was not using a nightly compiler.

I don't know if this change should actually be merged - do we always use the nightly rustdoc output?

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+# Pinning our toolchain ensures ./scripts/regenerate_test_rustdocs.sh always
+# emits the desired format.
+
+[toolchain]
+channel = "nightly"


obi1kenobi:
> do we always use the nightly rustdoc output?

Unfortunately, no. There are multiple different branches in this repo that are "live" concurrently, and each of them target a different rustdoc format version. Some of those versions are now stable and historical, some of them are current-stable, some of them are nightly-only. Some of them are currently nightly-only but will soon become stable.

The point of friction you identified is real and worth solving. Thanks for bringing it up!

Unfortunately, just mentioning `+nightly` might make it even more confusing, since it might not work. Any ideas on how we might make the suggestion better and more applicable? We also should avoid hardcoding any constants in the message that we can't test for and we need to remember to bump when new formats are released, because we are _practically guaranteed_ to forget.

fprasx:
Maybe we could use a `rust-toolchain`.toml. This can be easily `include_str!`d and the toolchain extracted:
```
fn get_toolchain() {
    toolchain = include_str(...);
    toolchain = for lines in toolchain, if line contains toolchain, get last whitespace separated word
    toolchain
}
```

On second thought, this only works if the rustdoc format only changes on release boundaries.

obi1kenobi:
Oh that's a good idea — if we set the expected Rust version in `rust-toolchain.toml`, won't it get picked up automatically when we run the command? We probably don't need to parse the file itself at all, it should "just work."

We can keep the most recent version on latest nightly (so we catch situations where it's no longer compatible or otherwise doesn't work) and for everything else we can pin a specific toolchain (nightly / beta / stable, as needed).

Wanna try it?

fprasx:
> won't it get picked up automatically when we run the command?

Right!

Although this will also mean that the adapter will have to compile for nightly (using normal `cargo run`).

fprasx:
Oh one thing: should I pin it to the current nightly, or just "nightly"?

EDIT: It seems like changes to the rustdoc JSON format are made through the `rust-lang/rust` repo. Therefore the format should only change when when a new nightly release is made. So we can just pin to the current nightly.

rust-toolchain.toml
```diff
@@ -0,0 +1,8 @@
+# This file pins the toolchain  so that ./scripts/regenerate_test_rustdocs.sh
+# always produces the desired format. Every new nightly release, if the rustdoc
+# format is unchanged, this toolchain may be incremented. Otherwise we need a new
+# version of the adapter.
+
+[toolchain]
+channel = "nightly-2024-01-09"
```
obi1kenobi: I'd maybe pin this to `nightly`, just to make sure we notice when a new format gets released and so we make sure things still work on that format -- e.g. in case of rustdoc bugs or bugfixes that don't change the format version. Those aren't super common, but they have happened and it's nice to catch them quickly.

fprasx: Fixed.

fprasx:
Hmm, pinning the toolchain seems to be causing CI to hang in the 'post install rust' step.

obi1kenobi:
Nice! Would you care to open PRs against the `rustdoc-v26` and `rustdoc-v27` branches with suitable `rust-toolchain.toml` files there as well? Those are the other two versions that are currently maintained.

PR #311, opened 2024-01-11
pin toolchain for rustdoc-v26

fprasx:


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+# Pinning our toolchain ensures ./scripts/regenerate_test_rustdocs.sh always
+# emits the desired format.
+
+[toolchain]
+channel = "nightly-2023-08-18"
```
obi1kenobi:
```suggestion
channel = "1.73"
```
This corresponds to stable Rust, so let's use the newest stable Rust that matches — it'll be mentioned in CI.

I wonder if this might cause a problem when CI tries to test this with multiple Rust versions 🤔

obi1kenobi: Filed https://github.com/actions-rust-lang/setup-rust-toolchain/issues/28 since CI ignores our explicit toolchain command and uses what's in the file — in the meantime, to unblock this PR, let's add a step to CI to delete the `rust-toolchain.toml` file before setting up Rust with that action.

fprasx: Yeah . . . should I hold off on PR'ing this?

obi1kenobi: Maybe put it on the back burner for a ~day or two. It seems like the GitHub action might get an update that would fix this behavior.

It might be good to open a draft PR for rustdoc-v27, just to make sure we don't forget to update it there. Then, assuming the action gets an update, I can update it in all three branches and both this PR and the rustdoc-v27 one would become directly mergeable.

fprasx: Sounds good.

obi1kenobi: The action got updated, so I'm going to re-trigger the tests here and I expect they should pass.

obi1kenobi: Hmm tests failed on output mismatch errors 🤔 Not sure what's going on.

Since it's your PR, I'll let you look into it unless you specifically want me to take over.

fprasx: I'll look into it tomorrow

fprasx: Resolving as stale. (new comments in discussion below)

fprasx:
This log line from `rustup` indicates that the action is not correctly overriding the toolchain file (this was from attempting to use `1.72`:
```
info: note that the toolchain '1.73-x86_64-unknown-linux-gnu' is currently in use 
(overridden by '.../rust-toolchain.toml')
```
which indicates that the action is not correctly overriding the toolchain file. `rustup override set $TOOLCHAIN` seems to work locally. Issue: https://github.com/actions-rust-lang/setup-rust-toolchain/issues/31

Confusingly though, the test cases are still failing. This seems related to `./scripts/regenerate_test_rustdocs.sh` being explicitly invoked with `+$TOOLCHAIN`, while cargo is not. Explicitly invoking cargo with `+$TOOLCHAIN` fixes things. I'm not sure why the adapter _compiled and run_ under `1.73` would be incompatible with rustdoc _generated_ by `1.72`.

It seems like we can wait for the action to merge a fix or merge this PR as is. Once we decide a course of action I'll pin the toolchain for `rustdoc-v27`.

obi1kenobi:
Let's merge this as-is. Since this is a branch that isn't the latest, I think it's probably not worth spending too much more energy on it.

Thanks for digging into the cause and opening the issue upstream, it looks good to me.

fprasx:
Hype. PR for `rustdoc-v27` coming up.

PR #312, opened 2024-01-12
pin toolchain for rustdoc-v27

fprasx:
pin toolchain to latest rustdoc-v27 stable (1.75)


obi1kenobi (approved):
Good to merge after #313 updates the Rust versions in CI.

Right now, it will try to test on `nightly` but current `nightly` uses rustdoc v28, not v27, so it won't work. I'll merge it shortly after #313 passes.

PR #315, opened 2024-01-13
add union types to schema (#309)

obi1kenobi:
* add union types to schema

* add union support to adapter

* avoid extra indirection when resolving union edge

* add union to supported item kinds

* resolve __typename of unions

This test case is also included in this commit as it exposed the bug.

* add unions to some glob test cases

* add tests


obi1kenobi:
I just realized, this test failure is caused by the bug reproduced in https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/300

I'm just going to stop testing on Rust 1.71 / 1.72, since the bug is fixed in 1.73+ and will never be backported to those older releases.

PR #323, opened 2024-02-10
feat: Object safe traits

devanbenz:
- Implements is_object_safe on Trait which determines whether a trait can be used as a trait object such as &dyn Trait or Box dyn Trait.


.gitignore
```diff
@@ -13,3 +13,6 @@ localdata/
 
 # lockfiles of test crates
 test_crates/**/Cargo.lock
+
+# ide
+.idea
```
obi1kenobi:
```suggestion
.idea

```

obi1kenobi (approved):
It was pretty easy, right? 😁

PR #337, opened 2024-07-21
Add enum discriminants

SuperSonicHub1:
Adds information about enum discriminants to the schema.

Refs:
- https://docs.rs/rustdoc-types/latest/rustdoc_types/struct.Variant.html
- https://docs.rs/rustdoc-types/latest/rustdoc_types/struct.Discriminant.html
- https://doc.rust-lang.org/reference/items/enumerations.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,5 @@
+name: CI
+
 on:
   pull_request:
   push:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        toolchain: ["1.77", "beta"]
+        toolchain: ["1.77", "1.78"]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        toolchain: ["1.77", "beta", "nightly"]
+        toolchain: ["1.77", "beta"]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -299,7 +299,7 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
 
             if let Some(discriminant) = &item.discriminant {
                 Box::new(std::iter::once(
-                    origin.make_discriminant_vertex(discriminant.clone()),
+                    origin.make_discriminant_vertex(discriminant),
                 ))
             } else {
                 Box::new(std::iter::empty())
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -1,5 +1,5 @@
 use rustdoc_types::{GenericBound::TraitBound, Id, ItemEnum, Variant, VariantKind};
-use std::sync::Arc;
+use std::rc::Rc;
 use trustfall::provider::{
     resolve_neighbors_with, AsVertex, ContextIterator, ContextOutcomeIterator, ResolveEdgeInfo,
     VertexIterator,
@@ -340,7 +340,7 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                         }
                     })
                     .collect();
-                Arc::new(LazyDiscriminants::new(variants))
+                Rc::new(LazyDiscriminants::new(variants))
             };
 
             Box::new(
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -298,9 +298,19 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
         }),
         "discriminant" => resolve_neighbors_with(contexts, move |vertex| {
             let origin = vertex.origin;
-            let enum_var = vertex.as_enum_variant().expect("vertex was not a Variant");
+            // HACK: to get around closure bounds
+            let enum_var = match &vertex.kind {
+                super::VertexKind::Variant(var) => var,
+                _ => unreachable!("vertex was not a Variant"),
+            };
+            let discriminant = enum_var
+                .discriminants
+                .get_discriminants()
+                .get(enum_var.index)
+                .unwrap()
+                .clone();
             Box::new(std::iter::once(
-                origin.make_discriminant_vertex(enum_var.discriminant().clone()),
+                origin.make_discriminant_vertex(discriminant),
             ))
         }),
         _ => unreachable!("resolve_variant_edge {edge_name}"),
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -1,4 +1,5 @@
-use rustdoc_types::{GenericBound::TraitBound, Id, ItemEnum, VariantKind};
+use rustdoc_types::{GenericBound::TraitBound, Id, ItemEnum, Variant, VariantKind};
+use std::sync::Arc;
 use trustfall::provider::{
     resolve_neighbors_with, AsVertex, ContextIterator, ContextOutcomeIterator, ResolveEdgeInfo,
     VertexIterator,
@@ -6,7 +7,9 @@ use trustfall::provider::{
 
 use crate::{adapter::supported_item_kind, attributes::Attribute, IndexedCrate};
 
-use super::{optimizations, origin::Origin, vertex::Vertex, RustdocAdapter};
+use super::{
+    enum_variant::LazyDiscriminants, optimizations, origin::Origin, vertex::Vertex, RustdocAdapter,
+};
 
 pub(super) fn resolve_crate_diff_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
     contexts: ContextIterator<'a, V>,
@@ -295,15 +298,10 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
         }),
         "discriminant" => resolve_neighbors_with(contexts, move |vertex| {
             let origin = vertex.origin;
-            let item = vertex.as_variant().expect("vertex was not a Variant");
-
-            if let Some(discriminant) = &item.discriminant {
-                Box::new(std::iter::once(
-                    origin.make_discriminant_vertex(discriminant),
-                ))
-            } else {
-                Box::new(std::iter::empty())
-            }
+            let enum_var = vertex.as_enum_variant().expect("vertex was not a Variant");
+            Box::new(std::iter::once(
+                origin.make_discriminant_vertex(enum_var.discriminant().clone()),
+            ))
         }),
         _ => unreachable!("resolve_variant_edge {edge_name}"),
     }
@@ -329,9 +327,35 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                         .index
                 }
             };
-            Box::new(enum_item.variants.iter().map(move |field_id| {
-                origin.make_item_vertex(item_index.get(field_id).expect("missing item"))
-            }))
+
+            let discriminants = {
+                let variants: Vec<&Variant> = enum_item
+                    .variants
+                    .iter()
+                    .map(move |field_id| {
+                        let inner = &item_index.get(field_id).expect("missing item").inner;
+                        match inner {
+                            ItemEnum::Variant(v) => v,
+                            _ => unreachable!("Item {inner:?} not a Variant"),
+                        }
+                    })
+                    .collect();
+                Arc::new(LazyDiscriminants::new(variants))
+            };
+
+            Box::new(
+                enum_item
+                    .variants
+                    .iter()
+                    .enumerate()
+                    .map(move |(index, field_id)| {
+                        origin.make_variant_vertex(
+                            item_index.get(field_id).expect("missing item"),
+                            discriminants.clone(),
+                            index,
+                        )
+                    }),
+            )
         }),
         _ => unreachable!("resolve_enum_edge {edge_name}"),
     }
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -293,6 +293,18 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                 })),
             }
         }),
+        "discriminant" => resolve_neighbors_with(contexts, move |vertex| {
+            let origin = vertex.origin;
+            let item = vertex.as_variant().expect("vertex was not a Variant");
+
+            if let Some(discriminant) = &item.discriminant {
+                Box::new(std::iter::once(
+                    origin.make_discriminant_vertex(discriminant.clone()),
+                ))
+            } else {
+                Box::new(std::iter::empty())
+            }
+        }),
         _ => unreachable!("resolve_variant_edge {edge_name}"),
     }
 }
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -306,7 +306,7 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                 .expect("vertex was not a Variant");
             let discriminant = enum_var.discriminant();
             Box::new(std::iter::once(
-                origin.make_discriminant_vertex(std::borrow::Cow::Borrowed(discriminant)),
+                origin.make_discriminant_vertex(discriminant),
             ))
         }),
         _ => unreachable!("resolve_variant_edge {edge_name}"),
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -1,4 +1,4 @@
-use rustdoc_types::{GenericBound::TraitBound, Id, ItemEnum, Variant, VariantKind};
+use rustdoc_types::{GenericBound::TraitBound, Id, ItemEnum, VariantKind};
 use std::rc::Rc;
 use trustfall::provider::{
     resolve_neighbors_with, AsVertex, ContextIterator, ContextOutcomeIterator, ResolveEdgeInfo,
@@ -266,7 +266,10 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
     match edge_name {
         "field" => resolve_neighbors_with(contexts, move |vertex| {
             let origin = vertex.origin;
-            let item = vertex.as_variant().expect("vertex was not a Variant");
+            let item = vertex
+                .as_variant()
+                .expect("vertex was not a Variant")
+                .variant();
             let item_index = match origin {
                 Origin::CurrentCrate => &current_crate.inner.index,
                 Origin::PreviousCrate => {
@@ -296,21 +299,14 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                 })),
             }
         }),
-        "discriminant" => resolve_neighbors_with(contexts, move |vertex| {
+        "discriminant" => resolve_neighbors_with(contexts, move |vertex: &'_ Vertex<'a>| {
             let origin = vertex.origin;
-            // HACK: to get around closure bounds
-            let enum_var = match &vertex.kind {
-                super::VertexKind::Variant(var) => var,
-                _ => unreachable!("vertex was not a Variant"),
-            };
-            let discriminant = enum_var
-                .discriminants
-                .get_discriminants()
-                .get(enum_var.index)
-                .unwrap()
-                .clone();
+            let enum_var = vertex
+                .as_variant()
+                .expect("vertex was not a Variant");
+            let discriminant = enum_var.discriminant();
             Box::new(std::iter::once(
-                origin.make_discriminant_vertex(discriminant),
+                origin.make_discriminant_vertex(std::borrow::Cow::Borrowed(discriminant)),
             ))
         }),
         _ => unreachable!("resolve_variant_edge {edge_name}"),
@@ -339,7 +335,8 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
             };
 
             let discriminants = {
-                let variants: Vec<&Variant> = enum_item
+                let len = enum_item.variants.len();
+                let variants = enum_item
                     .variants
                     .iter()
                     .map(move |field_id| {
@@ -350,7 +347,7 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                         }
                     })
                     .collect();
-                Rc::new(LazyDiscriminants::new(variants))
+                Rc::new(LazyDiscriminants::new(variants, len))
             };
 
             Box::new(
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -301,9 +301,7 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
         }),
         "discriminant" => resolve_neighbors_with(contexts, move |vertex: &'_ Vertex<'a>| {
             let origin = vertex.origin;
-            let enum_var = vertex
-                .as_variant()
-                .expect("vertex was not a Variant");
+            let enum_var = vertex.as_variant().expect("vertex was not a Variant");
             let discriminant = enum_var.discriminant();
             Box::new(std::iter::once(
                 origin.make_discriminant_vertex(discriminant),
@@ -335,7 +333,6 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
             };
 
             let discriminants = {
-                let len = enum_item.variants.len();
                 let variants = enum_item
                     .variants
                     .iter()
@@ -347,7 +344,7 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                         }
                     })
                     .collect();
-                Rc::new(LazyDiscriminants::new(variants, len))
+                Rc::new(LazyDiscriminants::new(variants))
             };
 
             Box::new(
--- a/src/adapter/enum_variant.rs
+++ b/src/adapter/enum_variant.rs
@@ -66,7 +66,7 @@ impl<'a> EnumVariant<'a> {
         }
     }
 
-    pub(super) fn discriminant(&'a self) -> Cow<'a, str> {
+    pub(super) fn discriminant(&self) -> Cow<'a, str> {
         self.discriminants
             .get_discriminants()
             .get(self.index)
--- a/src/adapter/enum_variant.rs
+++ b/src/adapter/enum_variant.rs
@@ -66,11 +66,12 @@ impl<'a> EnumVariant<'a> {
         }
     }
 
-    pub(super) fn discriminant(&'a self) -> &'a Cow<'a, str> {
+    pub(super) fn discriminant(&'a self) -> Cow<'a, str> {
         self.discriminants
             .get_discriminants()
             .get(self.index)
             .expect("self.index should exist in self.discriminants")
+            .clone()
     }
 
     #[inline]
--- a/src/adapter/enum_variant.rs
+++ b/src/adapter/enum_variant.rs
@@ -1,0 +1,304 @@
+use rustdoc_types::{Item, ItemEnum, Variant};
+use std::fmt;
+use std::num::ParseIntError;
+use std::sync::Arc;
+use std::{str::FromStr, sync::OnceLock};
+
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub(super) struct EnumVariant<'a> {
+    item: &'a Item,
+    discriminants: Arc<LazyDiscriminants<'a>>,
+    index: usize,
+}
+
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub(super) struct LazyDiscriminants<'a> {
+    variants: Vec<&'a Variant>,
+    discriminants: OnceLock<Vec<String>>,
+}
+
+impl<'a> LazyDiscriminants<'a> {
+    pub(super) fn new(variants: Vec<&'a Variant>) -> Self {
+        Self {
+            variants,
+            discriminants: OnceLock::new(),
+        }
+    }
+
+    pub(super) fn get_discriminants(&self) -> &Vec<String> {
+        self.discriminants
+            .get_or_init(|| assign_discriminants(&self.variants))
+    }
+}
+
+impl<'a> EnumVariant<'a> {
+    pub(super) fn new(
+        item: &'a Item,
+        discriminants: Arc<LazyDiscriminants<'a>>,
+        index: usize,
+    ) -> Self {
+        Self {
+            item,
+            discriminants,
+            index,
+        }
+    }
+
+    pub(super) fn variant(&self) -> Option<&'a Variant> {
+        match &self.item.inner {
+            ItemEnum::Variant(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub(super) fn discriminant(&'a self) -> &'a String {
+        self.discriminants
+            .get_discriminants()
+            .get(self.index)
+            .unwrap()
+    }
+}
+
+enum DiscriminantValue {
+    I64(i64),
+    U64(u64),
+    I128(i128),
+    U128(u128),
+}
+
+impl DiscriminantValue {
+    pub fn max(&self) -> bool {
+        matches!(self, DiscriminantValue::U128(u128::MAX))
+    }
+
+    pub fn increment(&self) -> DiscriminantValue {
+        match self {
+            DiscriminantValue::I64(i) => {
+                match i.checked_add_unsigned(1) {
+                    // No overflow
+                    Some(i) => i.into(),
+                    // Overflow, number will fit in a u64
+                    None => DiscriminantValue::from(u64::try_from(*i).unwrap()).increment(),
+                }
+            }
+            DiscriminantValue::U64(i) => {
+                match i.checked_add(1) {
+                    // No overflow
+                    Some(i) => i.into(),
+                    // Overflow, number will fit in a i128
+                    None => DiscriminantValue::from(i128::from(*i)).increment(),
+                }
+            }
+            DiscriminantValue::I128(i) => {
+                match i.checked_add_unsigned(1) {
+                    // No overflow
+                    Some(i) => i.into(),
+                    // Overflow, number will fit in a u128
+                    None => DiscriminantValue::from(u128::try_from(*i).unwrap()).increment(),
+                }
+            }
+            DiscriminantValue::U128(i) => (i + 1).into(),
+        }
+    }
+}
+
+impl fmt::Display for DiscriminantValue {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            DiscriminantValue::I64(i) => write!(f, "{}", i),
+            DiscriminantValue::U64(i) => write!(f, "{}", i),
+            DiscriminantValue::I128(i) => write!(f, "{}", i),
+            DiscriminantValue::U128(i) => write!(f, "{}", i),
+        }
+    }
+}
+
+impl From<i64> for DiscriminantValue {
+    fn from(value: i64) -> Self {
+        DiscriminantValue::I64(value)
+    }
+}
+
+impl From<i128> for DiscriminantValue {
+    fn from(value: i128) -> Self {
+        DiscriminantValue::I128(value)
+    }
+}
+
+impl From<u64> for DiscriminantValue {
+    fn from(value: u64) -> Self {
+        DiscriminantValue::U64(value)
+    }
+}
+
+impl From<u128> for DiscriminantValue {
+    fn from(value: u128) -> Self {
+        DiscriminantValue::U128(value)
+    }
+}
+
+impl FromStr for DiscriminantValue {
+    type Err = ParseIntError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Ok(i) = i64::from_str(s) {
+            return Ok(i.into());
+        }
+        if let Ok(i) = u64::from_str(s) {
+            return Ok(i.into());
+        }
+        if let Ok(i) = i128::from_str(s) {
+            return Ok(i.into());
+        }
+        match u128::from_str(s) {
+            Ok(i) => Ok(i.into()),
+            Err(e) => Err(e),
+        }
+    }
+}
+
+/// <https://doc.rust-lang.org/reference/items/enumerations.html#assigning-discriminant-values>
+pub(super) fn assign_discriminants(variants: &Vec<&Variant>) -> Vec<String> {
+    let mut last: DiscriminantValue = DiscriminantValue::I64(0);
+    let mut discriminants: Vec<String> = Vec::with_capacity(variants.len());
+    for v in variants {
+        discriminants.push(match &v.discriminant {
+            Some(d) => {
+                last = DiscriminantValue::from_str(&d.value).unwrap();
+                d.value.clone()
+            }
+            None => last.to_string(),
+        });
+        if !last.max() {
+            last = last.increment();
+        }
+    }
+    discriminants
+}
+
+#[cfg(test)]
+mod tests {
+    use rustdoc_types::{Discriminant, VariantKind};
+
+    use super::*;
+
+    #[test]
+    fn i64() {
+        let explicit_1 = Variant {
+            discriminant: Some(Discriminant {
+                value: "5".into(),
+                expr: "".into(),
+            }),
+            kind: VariantKind::Plain,
+        };
+        let explicit_2 = Variant {
+            discriminant: Some(Discriminant {
+                value: "7".into(),
+                expr: "".into(),
+            }),
+            kind: VariantKind::Plain,
+        };
+        let explicit_3 = Variant {
+            discriminant: Some(Discriminant {
+                value: "-59999".into(),
+                expr: "".into(),
+            }),
+            kind: VariantKind::Plain,
+        };
+        let variants = vec![
+            &Variant {
+                discriminant: None,
+                kind: VariantKind::Plain,
+            },
+            &Variant {
+                discriminant: None,
+                kind: VariantKind::Plain,
+            },
+            &explicit_1,
+            &explicit_2,
+            &Variant {
+                discriminant: None,
+                kind: VariantKind::Plain,
+            },
+            &explicit_3,
+            &Variant {
+                discriminant: None,
+                kind: VariantKind::Plain,
+            },
+        ];
+        let actual = assign_discriminants(&variants);
+        let expected: Vec<String> = vec![
+            "0".into(),
+            "1".into(),
+            "5".into(),
+            "7".into(),
+            "8".into(),
+            "-59999".into(),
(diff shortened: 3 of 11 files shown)


SuperSonicHub1:
Also, on why `repr` is not discussed in this PR:
- the `Discriminant` rustdoc type contains no info about `repr` :/
- `repr` is an attr, so that should probably be the responsibility of `Enum`, not `Discriminant`
- parsing the `repr` attr is complex, so it likely should be a separate issue
  - https://doc.rust-lang.org/nomicon/repr-rust.html
  - https://doc.rust-lang.org/nomicon/other-reprs.html
  - https://docs.rs/parenthesized_c/latest/parenthesized_c/

src/adapter/vertex.rs
```diff
@@ -36,6 +36,7 @@ pub enum VertexKind<'a> {
     ImplementedTrait(&'a Path, &'a Item),
     FunctionParameter(&'a str),
     FunctionAbi(&'a Abi),
+    Discriminant(Discriminant),
```
obi1kenobi: Any reason not to return `&'a Discriminant` and avoid cloning?
```suggestion
    Discriminant(&'a Discriminant),
```

src/adapter/vertex.rs
```diff
@@ -254,6 +256,13 @@ impl<'a> Vertex<'a> {
             _ => None,
         }
     }
+
+    pub(super) fn as_discriminant(&self) -> Option<rustdoc_types::Discriminant> {
+        match &self.kind {
+            VertexKind::Discriminant(discriminant) => Some(discriminant.clone()),
```
obi1kenobi: If you choose to make the vertex hold a reference, then this `.clone()` would be avoided:
```suggestion
            VertexKind::Discriminant(discriminant) => Some(discriminant),
```

src/rustdoc_schema.graphql
```diff
@@ -429,6 +429,15 @@ interface Variant implements Item {
...
   field: [StructField!]
+  discriminant: Discriminant
+}
+
+"""
+https://docs.rs/rustdoc-types/latest/rustdoc_types/struct.Discriminant.html
+"""
+type Discriminant {
+  expr: String!
+  value: String!
```
obi1kenobi: Could you document what each of these hold? Some of the less-obvious properties have docs, and these seem pretty non-obvious — especially given that they are both `String!`-typed, which seems unexpected to me.

src/adapter/tests.rs
```diff
@@ -1657,3 +1657,63 @@ fn unions() {
...
+
+    similar_asserts::assert_eq!(
+        vec![
+            Output {
+                expr: "1".into(),
+                value: "1".into(),
+            },
+            Output {
+                expr: "{ _ }".into(),
+                value: "2".into(),
```
obi1kenobi: Is `value` always guaranteed to hold a number, or can it be any string?

test_crates/enum_discriminants/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+#[repr(C)]
+pub enum A {
+    One = 1,
+    Two = 1 + 1,
+}
```
obi1kenobi, line 44: Consider adding some more test cases for discriminants that are defined implicitly.

For example:
```rust
pub enum Example {
    Zero,
    Hundred = 100,
    HundredAndOne,
}
```
has three defined discriminants, 0, 100, and 101, per the reference: https://doc.rust-lang.org/reference/items/enumerations.html#implicit-discriminants

Similarly, we want to check that implicit discriminants on tuple and struct enums work as well, as the reference allows those too if the enum has an explicit primitive repr.

obi1kenobi (commented):
I don't think we can get away from parsing the enum's `repr`, since the existence of an explicit `repr` sometimes causes discriminants to be defined implicitly. I suspect those implicit discriminants might not be part of rustdoc JSON, but I've suggested a test that can check that.

There's already some attribute-parsing code in the adapter, so you don't have to reinvent the wheel there.

src/adapter/vertex.rs
SuperSonicHub1: This is definitely better; was just rushing to get something working before the end of the hackathon and so didn't want to deal with lifetimes.

src/adapter/vertex.rs
SuperSonicHub1: See https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/337#discussion_r1685854142.

SuperSonicHub1:
To more generally address some of the previous comments due to me leaving stuff out of docs:
- rustdoc only lists explicit discriminations, so we'll have to do the calculations for implicit discriminations ourselves. I'll add this in a new patch; let me know if we should think on this instead. (I assume we're going to squash this PR. :))

- `value` is always a number, but it's in the range `[-(2^127), 2^128)` according to the docs, so I would agree with the rustdoc devs that it should stay a string. If this would be a limitation on cargo-semver-checks, we'll need to either add bigints or an i256 software impl to Trustfall.
> The numerical value of the discriminant. Stored as a string due to JSON’s poor support for large integers, and the fact that it would need to store from [i128::MIN](https://doc.rust-lang.org/nightly/std/primitive.i128.html#associatedconstant.MIN) to [u128::MAX](https://doc.rust-lang.org/nightly/std/primitive.u128.html#associatedconstant.MAX).

test_crates/enum_discriminants/src/lib.rs
SuperSonicHub1: Can do this for explicit enumerations. Explain more about implicit enumerations in a comment below.

src/adapter/tests.rs
SuperSonicHub1: It is always a number in a string AFAIK. See comment below.

src/rustdoc_schema.graphql
SuperSonicHub1: I explain in a comment below why rustdoc stores both, especially `value` as strings.

SuperSonicHub1:
> rustdoc only lists explicit discriminations, so we'll have to do the calculations for implicit discriminations ourselves.

I just remembered that the datatype of `value` is complicated due to `repr`, which I currently don't support. :)

I'll hold off on that until I hear back from you.

src/rustdoc_schema.graphql
```diff
@@ -429,6 +429,29 @@ interface Variant implements Item {
 
   # own edges
   field: [StructField!]
+
+  """
+  The discriminant, if explicitly specified.
+  """
+  discriminant: Discriminant
```
obi1kenobi: I think "if explicitly specified" is going to cause some trouble with writing lints.

For example, how would this schema support a lint that catches the following breakage:
```rust
#[derive(PartialOrd)]
pub enum Example {
    // to break: swap these two lines
    A,
    B,
}
```
This is breaking in two ways: the implicit discriminant changes, and the enum sorts differently than before.

What do you think about defining the `discriminant` edge as `The discriminant, if one is defined either explicitly or implicitly.` + a link to the Rust reference and possibly a couple of inline examples.

A possible representation for the `VertexKind::Discriminant` variant might be something like:
```rust
pub enum DiscriminantValue<'a> {
    Explicit(&'a Discriminant),
    Implicit(Arc<str>),  // holds the `value`, and always has `expr = "_"`
}
```

src/rustdoc_schema.graphql
```diff
@@ -429,6 +429,29 @@ interface Variant implements Item {
...
+
+"""
+https://docs.rs/rustdoc-types/0.28.0/rustdoc_types/struct.Discriminant.html
+https://doc.rust-lang.org/reference/items/enumerations.html
+"""
+type Discriminant {
+  """
+  The expression that produced the discriminant. Preserves the original formatting seen in source code (e.g. suffixes, hexadecimal, and underscores).
+
+  In some cases, when the value is to complex, this may be "`{ _ }`".
```
```suggestion
  In some cases, when the value is too complex, this may be a placeholder value such as `"{ _ }"`.
```
We cannot promise a specific placeholder value, because rustdoc doesn't promise that to us. It may at any point decide to switch to `"_"`.

Also note that I swapped the backtick and quote, since I think they were backwards in the original.

src/rustdoc_schema.graphql
```diff
@@ -429,6 +429,29 @@ interface Variant implements Item {
...
+  """
+  expr: String!
+
+  """
+  The numerical value of the discriminant. Stored as a string. Ranges from `i128::MIN` to `u128::MAX`.
```
obi1kenobi, line 460: Minor phrasing tweak suggestion:
```suggestion
  The numerical value of the discriminant, represented as a string.
  
  May range from `i128::MIN` to `u128::MAX`.
```

test_crates/enum_discriminants/src/lib.rs
obi1kenobi: Assuming you agree with my suggestion about changing the definition away from `"The discriminant, if explicitly specified."` to one where the edge exists if the discriminant is well-defined _via any mechanism_, it would be useful to have a test that we compute the implicit discriminants correctly in all cases.

For example:
```
pub enum Example {
    Zero,
    One,
    Two,
}
```
has well-defined discriminants: 0, 1, 2. They don't have to be specified explicitly here.

src/rustdoc_schema.graphql
obi1kenobi: A possible other extension to the schema, assuming you agree with the above, might be to add an `explicit: Boolean!` property to `type Discriminant`, for the sake of better diagnostic messages.

obi1kenobi (commented):
I like where this is going! Thanks for putting it together!

I think the major design question is what to do with implicitly-defined discriminants.

Since lints care about whether the discriminant's value is well-defined, and don't care about _how_ it's defined, I think it'll be more helpful for lints if we always provide the discriminant if it's well-known.

I made a case for this within.

obi1kenobi:
> I just remembered that the datatype of `value` is complicated due to `repr`, which I currently don't support. :)

I think it's probably fine to keep it as a string.

If you don't want to include the data type in the schema, you don't have to. I don't think it'll be strictly necessary to compute discriminant values, since they'll get turned into strings anyway.

It _might_ be worth trying to add a fast path that avoids 128-bit integer math, and uses 64-bit math as much as possible when computing implicit discriminants. 128-bit ints can be quite expensive, and almost no enums will actually use them in practice.

SuperSonicHub1:
> If you don't want to include the data type in the schema, you don't have to. I don't think it'll be strictly necessary to compute discriminant values, since they'll get turned into strings anyway.

I think `repr(..)` is a complex topic; I want to avoid overscoping this PR.

> It might be worth trying to add a fast path that avoids 128-bit integer math, and uses 64-bit math as much as possible when computing implicit discriminants. 128-bit ints can be quite expensive, and almost no enums will actually use them in practice.

Even if I had the type, parsing explicit discriminants is hard because I don't know the type until I a number becomes too large or too small for the type I presumed. Below is what I believe is the minimum code for representing any possible discriminant value. I believe use of `unwrap` is justified as rustc has already made sure that all enum discriminants fit inside either an `u128` or an `i128`. Let me know if it looks good.

```rust
enum DiscriminantValue {
    I64(i64),
    U64(u64),
    I128(i128),
    U128(u128),
}

impl DiscriminantValue {
    fn increment(self) -> DiscriminantValue {
        match self {
            DiscriminantValue::I64(i) => {
                match i.checked_add_unsigned(1) {
                    Some(i) => i.into(),
                    None => {
                        match u64::try_from(i) {
                            Ok(u) => DiscriminantValue::from(u).increment(),
                            // i is a negative i128
                            Err(_) => DiscriminantValue::from(i128::try_from(i).unwrap()).increment(),
                        }
                    },
                }
            },
            DiscriminantValue::U64(i) => {
                match i.checked_add(1) {
                    Some(i) => i.into(),
                    None => {
                        match i128::try_from(i) {
                            Ok(u) => DiscriminantValue::from(u).increment(),
                            // i is a u128
                            Err(_) => DiscriminantValue::from(u128::try_from(i).unwrap()).increment(),
                        }
                    },
                }
            },
            DiscriminantValue::I128(i) => {
                match i.checked_add_unsigned(1) {
                    Some(i) => i.into(),
                    // i is a u128
                    None => DiscriminantValue::from(u128::try_from(i).unwrap()).increment(),
                }
            },
            DiscriminantValue::U128(i) => (i + 1).into(),
        }
    }
}

impl From<i64> for DiscriminantValue {
    fn from(value: i64) -> Self {
        DiscriminantValue::I64(value)
    }
}

impl From<i128> for DiscriminantValue {
    fn from(value: i128) -> Self {
        DiscriminantValue::I128(value)
    }
}

impl From<u64> for DiscriminantValue {
    fn from(value: u64) -> Self {
        DiscriminantValue::U64(value)
    }
}

impl From<u128> for DiscriminantValue {
    fn from(value: u128) -> Self {
        DiscriminantValue::U128(value)
    }
}

impl FromStr for DiscriminantValue {
    type Err = ParseIntError;

    
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        if let Ok(i) = i64::from_str(s) {
            return Ok(i.into());
        }
        if let Ok(i) = u64::from_str(s) {
            return Ok(i.into());
        }
        if let Ok(i) = i128::from_str(s) {
            return Ok(i.into());
        }
        match u128::from_str(s) {
            Ok(i) => Ok(i.into()),
            Err(e) => Err(e),
        }
    }
}
```

obi1kenobi:
I have a few concerns that could be resolved by seeing more of how you plan to integrate that enum into the broader design:
- I'm wary of including lots of enum nesting each with only a few variants at a time. Each time we nest enums, we pay for a fresh discriminant which costs memory, which in turn costs performance.
- Relatedly, that enum is quite large (at least 17 bytes) and in the vast majority of cases will contain 16 bytes of zeroes. That will also have a performance cost, especially if we include it in the vertex.
- I think at least one branch in the shown `increment()` function is actually unreachable: an `i64` being incremented by 1 must always fit either in `i64` or `u64`. It cannot require `i128` because it cannot become any more negative, nor exceed u64::MAX while starting from i64::MAX. More careful analysis would be useful to minimize the amount of code there.

SuperSonicHub1:
Spent some time while on the move simplifying `increment` and sketching out discriminant assignment: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f61c7eb6a08c32afee8d13a88a36f6f6

Test looks like it demonstrates expected behavior; perhaps should include more pathological results like i128 and u64 instances. If you think of a more functional and less mutable way of calculating the discriminants before I do, let me know. I'll more neatly integrate the algo into the codebase in terms of lifetimes once I'm at my desk again.

I think I'm going to calculate the discriminants all at once  when the `variants` edge is requested. The algorithm is linear, so it shouldn't take very long for non-pathological codebases. Passing around a `&'a String` with a variant's `Id` shouldn't be too bad, I think.

Finally, I have decided that I'm going to remove `Discriminant::expr` from the schema; it's really only present for cosmetic reasons relevant to Rustdoc and not us, and has no use in any of the lints you have planned. As a wise mentor (you) once told me long ago, "only implement what you need."

obi1kenobi:
This looks quite reasonable, thank you!

I'd recommend a couple of small-ish tweaks, in addition to recommending more test coverage with i128 and u128 discriminants before we merge.

Consider either `increment(&self)` or `increment(&mut self)` instead of `increment(self)`? The underlying type is `Copy` so the move doesn't look to me like it's buying us very much.

Also, consider (either in this PR or in a subsequent one) making it so that the discriminants are only computed when a _discriminant_ edge is requested, not a _variant_ edge? This is because there will be lots of lints that touch variants without touching discriminants, and we'd prefer to be as lazy as possible here. Ideally, we'd also memoize the computed discriminant strings if possible so we don't compute them multiple times if we have multiple lints over them.

The outcome we'd like to avoid in the long run is having many lints repeatedly compute discriminant strings (and do a bunch of allocations in the process) even though very few of them actually ever need discriminants.

SuperSonicHub1:
> Also, consider (either in this PR or in a subsequent one) making it so that the discriminants are only computed when a _discriminant_ edge is requested, not a _variant_ edge? This is because there will be lots of lints that touch variants without touching discriminants, and we'd prefer to be as lazy as possible here. 

I was thinking about how to do the former the other day; the Rustdoc just makes it really annoying to get a `Variant`'s parent `Enum` and so its sibling enums. This is necessary as a variant's discriminant conditionally depends on the variants before it.

> Ideally, we'd also memoize the computed discriminant strings if possible so we don't compute them multiple times if we have multiple lints over them.

Does Trustfall have some kind of general caching utility? If not, caching for our use case could be easily achieved with `once_cell`.

obi1kenobi:
> I was thinking about how to do the former the other day; the Rustdoc just makes it really annoying to get a `Variant`'s parent `Enum` and so its sibling enums. This is necessary as a variant's discriminant conditionally depends on the variants before it.
>
> Does Trustfall have some kind of general caching utility? If not, caching for our use case could be easily achieved with `once_cell`.

Here's what I'd do.

Right now, variants are represented as `VertexKind::Item`. I'd consider carving out a separate variant, `VertexKind::Variant(EnumVariant<'a>)` for them, where `EnumVariant` looks roughly like this:
```rust
#[non_exhaustive]
#[derive(Debug, Clone)]
pub struct EnumVariant<'a> {
    item: &'a Item,
    discriminant: todo!("< something clever that holds the parent `&'a Enum` internally and is lazily initialized >")
}
```

Then on first access to the `discriminant` field, we can compute all the discriminants for all the variants of that enum. This would imply some `Rc` inside there + a variant-specific offset or another way to indicate which of the bulk-computed discriminants corresponds to the current variant.

This sidesteps the problem of getting the `&'a Enum` for a variant, since we construct the `EnumVariant<'a>` value while we still know the `&'a Item` + `&'a Enum` that own the variant — we're coming from that edge.

SuperSonicHub1:
Realized that in order to do the first thing, I'll be 90% of the way to doing the other, so might as well do it all in one patchset.

SuperSonicHub1:
After much fighting with `rustc` and two afternoons of rumination, I got some working tests. Please suggest ways for me to clone less and be more memory efficient.

Will add `OrderedVariant` (do you think simply adding `index` to the Variant type would be better? I do) after a much deserved break.

SuperSonicHub1:
Messed up CI; on that.

SuperSonicHub1:
Aside: getting pre-commit hooks going for `cargo clippy && cargo fmt && cargo doc && cargo test` would be appreciated.

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,304 @@
...
+use std::fmt;
+use std::num::ParseIntError;
+use std::sync::Arc;
+use std::{str::FromStr, sync::OnceLock};
+
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub(super) struct EnumVariant<'a> {
+    item: &'a Item,
+    discriminants: Arc<LazyDiscriminants<'a>>,
```
obi1kenobi: We don't actually care about `Send` bounds, and we use `Rc` everywhere else in the adapter code. `cargo-semver-checks` runs queries to completion on the same thread only. 
```suggestion
    discriminants: Rc<LazyDiscriminants<'a>>,
```

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,304 @@
...
+    index: usize,
+}
+
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub(super) struct LazyDiscriminants<'a> {
+    variants: Vec<&'a Variant>,
+    discriminants: OnceLock<Vec<String>>,
```
obi1kenobi: Similarly, this doesn't need to be a lock since we'll only be using it from one thread.

src/adapter/edges.rs
```diff
@@ -317,9 +327,35 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
...
+                    .variants
+                    .iter()
+                    .map(move |field_id| {
+                        let inner = &item_index.get(field_id).expect("missing item").inner;
+                        match inner {
+                            ItemEnum::Variant(v) => v,
+                            _ => unreachable!("Item {inner:?} not a Variant"),
+                        }
+                    })
+                    .collect();
```
obi1kenobi, line 346: Can we avoid this `.collect()` somehow, and instead pass some flavor of `impl Iterator<&'a Variant> + 'a` into `LazyDiscriminants`?

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,304 @@
...
+
+/// <https://doc.rust-lang.org/reference/items/enumerations.html#assigning-discriminant-values>
+pub(super) fn assign_discriminants(variants: &Vec<&Variant>) -> Vec<String> {
+    let mut last: DiscriminantValue = DiscriminantValue::I64(0);
+    let mut discriminants: Vec<String> = Vec::with_capacity(variants.len());
+    for v in variants {
+        discriminants.push(match &v.discriminant {
+            Some(d) => {
+                last = DiscriminantValue::from_str(&d.value).unwrap();
+                d.value.clone()
```
obi1kenobi: If we make `discriminants` be a `Vec<Cow<'a, str>>`, I think we can avoid the clone here. This will be `Cow::Borrowed(&d.value)` and the other arm will be `Cow::Owned(last.to_string())`.

Another option which is also tempting is to adapt to Trustfall's `FieldValue::String` representation of `Arc<str>` for owned strings, and have an enum holding either a borrowed `&'a str` or an `Arc<str>`, rather than the `String` owned variant.

The latter is probably going to be a bit more advantageous perf-wise, but it's also a bit more work and you don't have to do it if you think it's scope-creep.

src/adapter/vertex.rs
```diff
@@ -36,31 +36,34 @@ pub enum VertexKind<'a> {
...
-                    VariantKind::Tuple(..) => "TupleVariant",
-                    VariantKind::Struct { .. } => "StructVariant",
-                },
                 rustdoc_types::ItemEnum::StructField(..) => "StructField",
                 rustdoc_types::ItemEnum::Impl(..) => "Impl",
                 rustdoc_types::ItemEnum::Trait(..) => "Trait",
                 rustdoc_types::ItemEnum::Constant(..) => "Constant",
                 rustdoc_types::ItemEnum::Static(..) => "Static",
                 rustdoc_types::ItemEnum::AssocType { .. } => "AssociatedType",
+                // TODO: How are we even here???
```
obi1kenobi: Was this a note-to-self, or something you wanted me to answer?

obi1kenobi (commented):
This is looking pretty good!

I made some suggestions on how to avoid some cloning and eager evaluations to further improve performance. Let me know if you'd like to pair-program any of this — sometimes working with iterators and lifetimes gets complicated, and I've debugged a lot of those errors so I'm happy to help.

Re: pre-commit hooks, I actually really hate tools that prevent me from committing since I do a lot of quick-and-dirty experimentation. But if you wanted to set it up for yourself in a way that doesn't impose a mandatory workflow change on other contributors, I'm fine with that.

src/adapter/vertex.rs
SuperSonicHub1: Both. I was having a really hard time understanding how I was ending up in this branch just reading `edges.rs`.

src/adapter/vertex.rs
```diff
@@ -36,31 +36,34 @@ pub enum VertexKind<'a> {
     ImplementedTrait(&'a Path, &'a Item),
     FunctionParameter(&'a str),
     FunctionAbi(&'a Abi),
+    Discriminant(String),
```
obi1kenobi:
```suggestion
    Discriminant(String),
```
I'd recommend either `Cow<'a, str>` or, even better, the borrowed-or-`Arc<str>` alternative. Vertices get cloned _a lot_, and we don't want to keep cloning the full `String` each time.

src/adapter/vertex.rs
obi1kenobi: This query will get you there:
```graphql
{
    Crate {
        item {
            ... on Enum {
                variant {
                    __typename @output
                }
            }
        }
    }
}
```

If the enum has all three kinds of variants, all three branches of the inner `match` will get hit too.

obi1kenobi: Also, remember that `match` arms are evaluated in order, so it's usually a good idea to not reorder them unless you've measured perf and determined the change is an improvement.

src/adapter/enum_variant.rs
SuperSonicHub1: I'll stick to the former for scope.

SuperSonicHub1:
Finally did it.
Reason for the `HACK: to get around closure bounds seen in src/adapter/edges.rs:301`:
```
33 | /         self.discriminants
34 | |             .get_or_init(|| assign_discriminants(&self.variants))
   | |_________________________________________________________________^ expected `&Vec<String>`, found `&Vec<Cow<'_, str>>`
   |
   = note: expected reference `&Vec<std::string::String>`
              found reference `&Vec<Cow<'a, str>>`

error[E0521]: borrowed data escapes outside of closure
   --> src/adapter/edges.rs:301:28
    |
299 |         "discriminant" => resolve_neighbors_with(contexts, move |vertex| {
    |                                                                  ------
```
Couldn't wrap my head around why the data was escaping, so I inlined the whole codepath.
Still working on passing an iterator to `assign_discriminants`.

SuperSonicHub1:
Also, sorry for the uneven spurts of progress; I know that can be a burden on maintainers.

obi1kenobi:
> Also, sorry for the uneven spurts of progress; I know that can be a burden on maintainers.

No worries, it's not like I was awash in spare time either this past week.

I'll try to poke at this and figure out what's with the borrowing problem. I don't see what was escaping from the closure, and I'm curious.

src/adapter/vertex.rs
```diff
@@ -165,10 +173,17 @@ impl<'a> Vertex<'a> {
...
-        self.as_item().and_then(|item| match &item.inner {
-            rustdoc_types::ItemEnum::Variant(v) => Some(v),
+        match &self.kind {
+            VertexKind::Variant(variant) => variant.variant(),
             _ => None,
-        })
+        }
+    }
+
+    pub(super) fn as_enum_variant(&self) -> Option<&'a EnumVariant> {
```
obi1kenobi: This signature is incorrect, and is causing the borrowing/escaping error you were dealing with:
```suggestion
    pub(super) fn as_enum_variant(&self) -> Option<&'_ EnumVariant<'a>> {
```

We can't create a `&'a EnumVariant` from an owned `VertexKind::Variant(EnumVariant<'a>)`. We can only create a `&'_ EnumVariant<'a>` so that's what we should return. Note how everything that returns a `&'a T` here actually held a `&'a T` internally, whereas everything else (e.g. `Attribute`) returns a `&'_ T`.

As an additional note, it's generally a bad idea to elide lifetime arguments from the type, like the `&'a EnumVariant` here instead of `&'a EnumVariant<'a>`. If we had the latter here, the issue would have been much more obvious.

src/adapter/vertex.rs
```diff
@@ -165,10 +173,17 @@ impl<'a> Vertex<'a> {
...
 
     pub(super) fn as_variant(&self) -> Option<&'a Variant> {
-        self.as_item().and_then(|item| match &item.inner {
-            rustdoc_types::ItemEnum::Variant(v) => Some(v),
+        match &self.kind {
+            VertexKind::Variant(variant) => variant.variant(),
             _ => None,
-        })
+        }
+    }
```
obi1kenobi: Consider refactoring the code so that we don't have two confusing methods that do almost-but-not-quite the same thing? We probably want `as_variant()` to return `Option<&'_ EnumVariant<'a>>` and then to refactor the code that needed the old signature.

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,308 @@
...
+            index,
+        }
+    }
+
+    pub(super) fn variant(&self) -> Option<&'a Variant> {
+        match &self.item.inner {
+            ItemEnum::Variant(v) => Some(v),
+            _ => None,
+        }
+    }
```
obi1kenobi: We probably want to return `&'a Variant` here, and panic in the `None` branch with a suitable message. We should never have a situation where an `EnumVariant` is constructed from a non-`Variant` item.

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,308 @@
...
+use std::cell::OnceCell;
+use std::fmt;
+use std::num::ParseIntError;
+use std::rc::Rc;
+use std::str::FromStr;
+
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub(super) struct EnumVariant<'a> {
+    item: &'a Item,
```
obi1kenobi, line 12: We might want an accessor function for this item:

```rust
#[inline]
pub(super) fn item(&self) -> &'a Item {
    self.item
}
```

src/adapter/vertex.rs
```diff
@@ -36,14 +36,16 @@ pub enum VertexKind<'a> {
     ImplementedTrait(&'a Path, &'a Item),
     FunctionParameter(&'a str),
     FunctionAbi(&'a Abi),
+    Discriminant(Cow<'a, str>),
+    Variant(EnumVariant<'a>),
```
obi1kenobi, line 40: `EnumVariant` might have to be `pub`. Let's keep it in sync visibility-wise with other types exposed in variants here, like `ImportablePath` and `AttributeMetaItem`.

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,308 @@
...
+            _ => None,
+        }
+    }
+
+    // HACK: to get around closure bounds seen in ./edges.rs:301
+    // pub(super) fn discriminant(&'a self) -> &'a Cow<'a, str> {
+    //     self.discriminants
+    //         .get_discriminants()
+    //         .get(self.index)
+    //         .unwrap()
```
obi1kenobi: Prefer `.expect()` over `.unwrap()` basically always. It really sucks to have to figure out what went wrong in a panic that doesn't explain what invariant was violated.

obi1kenobi (commented):
I found the problem, see below 👇

It might also be a good idea to open a diagnostic bug report on rustc about the confusing diagnostic message from this borrow check error. It took me a while to find it, and I already have a decent bit of experience giving me a solid idea of where to look. Esteban and other folks working on compiler diagnostics would appreciate details on this, and I'm sure they'd want to take a look.

src/adapter/vertex.rs
SuperSonicHub1: Thanks for the catch; didn't even realize I elided!

obi1kenobi:
No pressure to move this forward, just wanted to mention two recent developments that are related.

I made a Discord server for `cargo-semver-checks` and related tools, to make it easier to coordinate the work. Here's an invite link good for the next 7 days, if you'd like to join: https://discord.gg/s7Cfmp3P

Also, there's been a recent influx of new contributors to the project. So if for any reason you don't have the time or interest to keep working on this PR, it's no trouble to find another person to complete it. I know what college is like and I know you have 20 different places to allocate your time, so I don't want you to feel like this PR is a burden on you. I'm in no rush, so you're welcome to keep working on it if you have a time frame for it in mind already.

SuperSonicHub1:
Thanks for checking in! Should have tons of time to code on Thursday, Friday, and the weekend to try to make the final pushes. Please do feel free to give my work to someone more eager!

obi1kenobi:
No worries, it's all yours! If you change your mind or if anything comes up, just let me know — it's no trouble.

SuperSonicHub1:
Still fighting the good fight against the compiler. Even with `Vertex::as_variant` now returning `Option<&'_ EnumVariant<'a>>`, I'm still running into the same kind of bounds issue; see the commit message.

I've found that I'm able to get more helpful compiler errors by making the elision of the reference's lifetime explicit in the closure (from `|vertex|` to `|vertex: &'_ Vertex<'a>|`).

SuperSonicHub1:
Still have no idea how to architect passing an iterator to `assign_discriminants`; do you have any code you recommend I study?

obi1kenobi:
As usual, rustc is right even though it isn't obvious why and how.

```rust
    pub(super) fn discriminant(&'a self) -> &'a Cow<'a, str> {
        self.discriminants
            .get_discriminants()
            .get(self.index)
            .expect("self.index should exist in self.discriminants")
    }
```
The `&'a self` there (almost always a red flag by itself btw!) means that this block:
```rust
            let enum_var = vertex
                .as_variant()
                .expect("vertex was not a Variant");
            let discriminant = enum_var.discriminant();
            Box::new(std::iter::once(
                origin.make_discriminant_vertex(std::borrow::Cow::Borrowed(discriminant)),
            ))
```
ends up moving `discriminant` AKA `&'1 Cow<'1 str>` into the iterator and out of the closure, where `&'1` is from `&'1 EnumDiscriminant<'a>` which in turn is from `&'1 Vertex<'a>`.

Follow it step-by-step: `enum_var` is `&'1 EnumDiscriminant<'a>` by `vertex.as_variant()`. Then `enum_var.discriminant()` says "the output lifetime is the same as the `&self` lifetime" and ends up applying `&'1` instead of `&'a` to the `Cow`. `'1` is shorter than `'a` and `Cow` is read-only, so this is fine and rustc allows it. Recall that inside `EnumDiscriminant<'a>`, that `'a` isn't the same as the `Vertex<'a>` one. This ends up being a problem.

obi1kenobi:
The following compiles fine:
```diff
diff --git a/src/adapter/edges.rs b/src/adapter/edges.rs
index ae57eed..9e3d3af 100644
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -306,7 +306,7 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                 .expect("vertex was not a Variant");
             let discriminant = enum_var.discriminant();
             Box::new(std::iter::once(
-                origin.make_discriminant_vertex(std::borrow::Cow::Borrowed(discriminant)),
+                origin.make_discriminant_vertex(discriminant),
             ))
         }),
         _ => unreachable!("resolve_variant_edge {edge_name}"),
diff --git a/src/adapter/enum_variant.rs b/src/adapter/enum_variant.rs
index bb89f5d..9d26d29 100644
--- a/src/adapter/enum_variant.rs
+++ b/src/adapter/enum_variant.rs
@@ -66,11 +66,12 @@ impl<'a> EnumVariant<'a> {
         }
     }
 
-    pub(super) fn discriminant(&'a self) -> &'a Cow<'a, str> {
+    pub(super) fn discriminant(&self) -> Cow<'a, str> {
         self.discriminants
             .get_discriminants()
             .get(self.index)
             .expect("self.index should exist in self.discriminants")
+            .clone()
     }
```

It's been a while so I don't actually recall why this is a `Cow` though. The `.clone()` call is cheap on `Cow::Borrowed` but expensive on `Cow::Owned` so we'd prefer to avoid that. And needing `&Cow` is usually a red flag that says something weird is going on too. So maybe we've made a subtle design error somewhere?

Anyway, you have more context about this PR right now than I do so hopefully you're able to piece it together from there!

SuperSonicHub1:
Ah, I'm a dolt for not cloning; that's the whole point of `Cow`!
To bring back [what you said](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/337#discussion_r1693996273), we're using `Cow` to have cheaper clones for our vector of discriminant values:
> If we make discriminants be a Vec<Cow<'a, str>>, I think we can avoid the clone here. This will be Cow::Borrowed(&d.value) and the other arm will be Cow::Owned(last.to_string()).

obi1kenobi:
Hmm cloning the owned arm might still be expensive. I'm not near a computer
right now, any good ways to avoid that?

Perhaps there's a way to call that `to_string()` later or something?

SuperSonicHub1:
> Perhaps there's a way to call that `to_string()` later or something?

Mayhaps.

I implemented your changes, but am still getting the same compiler error. I think I'll step away from the computer for a bit and think about what I'm doing wrong.

obi1kenobi:
You still have a `&'a self` which isn't present in my diff, I think that's
the difference.

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,323 @@
...
+            Ok(i) => Ok(i.into()),
+            Err(e) => Err(e),
+        }
+    }
+}
+
+/// <https://doc.rust-lang.org/reference/items/enumerations.html#assigning-discriminant-values>
+pub(super) fn assign_discriminants<'a>(variants: &Vec<&'a Variant>, len: usize) -> Vec<Cow<'a, str>> {
+    let mut last: DiscriminantValue = DiscriminantValue::I64(0);
+    let mut discriminants: Vec<Cow<'a, str>> = Vec::with_capacity(len);
```
obi1kenobi: I'm confused why we're threading this `len` everywhere when it appears to be the same thing as `variants.len()`.

This capacity is exactly known to be the same as `variants.len()` because we for-loop over `variants` and push on each iteration.

src/adapter/enum_variant.rs
```diff
@@ -0,0 +1,323 @@
...
+        }
+        match u128::from_str(s) {
+            Ok(i) => Ok(i.into()),
+            Err(e) => Err(e),
+        }
+    }
+}
+
+/// <https://doc.rust-lang.org/reference/items/enumerations.html#assigning-discriminant-values>
+pub(super) fn assign_discriminants<'a>(variants: &Vec<&'a Variant>, len: usize) -> Vec<Cow<'a, str>> {
```
obi1kenobi: It's a bit more idiomatic and efficient to pass `&[T]` (wide-pointer-to-data) than `&Vec<T>` (pointer-to-wide-pointer-to-data). The call site shouldn't have to do anything, the coercion should happen automatically.
```suggestion
pub(super) fn assign_discriminants<'a>(variants: &[&'a Variant], len: usize) -> Vec<Cow<'a, str>> {
```

obi1kenobi (commented):
I looked through the use of `Cow::Owned` and it's going to be too much work to make that path faster for now. Feel free to leave it as-is and we can tackle it in the future. We'll probably need to make our own `Cow` type where the owned variant is `Arc<str>` instead of `String` to make that clone cheap too. So let's just leave a TODO / open an issue and leave it for later.

I'm generally just confused by the `len` we threaded through everywhere that seems to be available already. Otherwise I'm happy to merge and release this.

src/adapter/enum_variant.rs
SuperSonicHub1: Ah, this is in preparation for passing an iterator to `assign_discriminants`; want the capacity of the returned array to be exactly as large as needed.

src/adapter/enum_variant.rs
SuperSonicHub1: Temporary for now; see above comment. If we want to give up on passing an iterator to `assign_discriminants` for now, I can do that.

src/adapter/enum_variant.rs
obi1kenobi: I'm worried passing the iterator will require a bigger overhaul, because of this:
```rust
    pub(super) fn get_discriminants(&self) -> &Vec<Cow<'a, str>> {
        self.discriminants
            .get_or_init(|| assign_discriminants(&self.variants, self.len))
    }
```

Using the iterator there will require `&mut self`, but this type is inside an `Rc` so we won't be able to get `&mut self`.

More improvements would be great, but I feel like this PR has gotten tortured enough 😅 So maybe we just merge it with the extra `.collect()` and leave a TODO + issue to look into fixing it if it becomes a problem?

SuperSonicHub1:
We'll come back to ordered variants another time.

SuperSonicHub1:
Otherwise, the patch should be done now!

src/adapter/tests.rs
```diff
@@ -1657,3 +1657,131 @@ fn unions() {
...
+    let adapter = RustdocAdapter::new(&indexed_crate, None);
+
+    let query = r#"
+{
+    Crate {
+        item {
+            ... on Enum {
+                name @output
+                variant {
+                    discriminant {
```
obi1kenobi: Would you mind outputting the variant names here, so we're positive we're correctly matching up variants and discriminants?

If at any point we happened to introduce a bug that makes that not be the case, finding it would be a horrifying debugging experience to go through.

src/rustdoc_schema.graphql
```diff
@@ -429,6 +429,22 @@ interface Variant implements Item {
 
   # own edges
   field: [StructField!]
+
+  """
+  The discriminant, if explicitly specified.
```
obi1kenobi: Here's a better doc:
```suggestion
  The discriminant value of this variant, if its value is well-defined.
  
  A discriminant is an integer value that is logically associated with an enum variant,
  and is used to determine which variant an enum value holds.

  Discriminant values are not always well-defined.

  In order for variants to have well-defined discriminants, their enum must either
  - define a primitive representation (e.g., `#[repr(u8)]`), or
  - be "unit-only" i.e. consisting solely of unit variants.

  The variants of such enums may be explicitly assigned a discriminant value, or
  the discriminant may be defined implicitly:
  - The first variant has a discriminant of zero, unless explicitly assigned a value.
  - Subsequent variants have a discriminant one higher than the previous variant,
    unless explicitly assigned a value.
```

test_crates/enum_discriminants/src/lib.rs
obi1kenobi, line 1: Can you add some test cases to show that when discriminants are *not* well-defined, we don't output them?

In the test query, you can use `discriminant @optional` to make that edge optional and have its outputs return `null` (deserialized to `Option::None`) when it doesn't exist.

src/rustdoc_schema.graphql
```diff
@@ -483,6 +499,7 @@ type PlainVariant implements Item & Variant {
 
   # edges from Variant
   field: [StructField!]
+  discriminant: Discriminant
```
obi1kenobi, line 536: Let's make sure to mirror the doc comment for all these edges.

Unfortunately, GraphQL LSP isn't smart enough to fall back to the supertype's definition of an edge if the concrete type's definition doesn't offer a doc comment. Duplicating is annoying, but it's less annoying than trying to use something that doesn't have docs.

obi1kenobi (commented):
Nice! Just some test improvements and polish items left. There appears to be a merge conflict too, which should be an easy fix.

I've retargeted the PR to branch `rustdoc-v33` instead of `rustdoc-v28`, so that we merge into the current development branch. That way I can easily port this to older versions, since I have some automation to handle that.

SuperSonicHub1:
I believe this should be it.

SuperSonicHub1:
Also, CI's been acting a bit weird these past few days; would you know why?

obi1kenobi:
Still seeing merge conflicts on my end, can you try pulling the latest upstream `rustdoc-v33` branch and merging it into your branch?

GitHub doesn't trigger CI runs on PRs that are unmergeable due to conflicts.

SuperSonicHub1:
> GitHub doesn't trigger CI runs on PRs that are unmergeable due to conflicts.

D'oh. Think I did it.

.github/workflows/ci.yml
```diff
@@ -1,5 +1,3 @@
-name: CI
-
```
obi1kenobi: This looks like it might have accidentally gotten deleted in the merge.

SuperSonicHub1: Thanks for the catch!

obi1kenobi (approved):
Excellent, thank you!

Any interest in continuing to work on variant order / struct field order? Or too busy with schoolwork? No pressure — cargo-semver-checks will be here if/when you want to come back to contributing — I just wanted to know whether you wanted dibs on those extensions or if I should put them up for grabs.

SuperSonicHub1:
Woo!

SuperSonicHub1:
> Any interest in continuing to work on variant order / struct field order?

I would, but school and research is likely going to be tough, so put them up for grabs. Plan for this summer is to find a way to get paid to contribute to OSS, so this will not be the last time you hear from me for the next 12 months. :) I also will have a ton of free time during winter break, so we should discuss possibilities then, too.

obi1kenobi:
If there are companies or programs you're interested in applying to, I'd be happy to serve as a reference / write you a recommendation letter / refer you if I know someone there. (Same goes for anyone who consistently contributes quality code to projects I maintain!)

I'm a mentor for Google Summer of Code, and from my perspective that program has been good! Not sure how companies / hiring managers feel about it on a resume (as opposed to say a tech company internship) so it's worth looking into that as well.

Also happy to work together over winter break — ping me whenever it's a good time to chat about that!

PR #389, opened 2024-08-19
feat: make has_body queryable on methods

mrnossiom:
Hi 👋🏻,

Retrieve the `has_body` property on function like items and expose it to queries.

If I understand correctly, this is what's blocking for some lints in obi1kenobi/cargo-semver-checks#870.

### DIFF
--- a/src/adapter/tests.rs
+++ b/src/adapter/tests.rs
@@ -1818,8 +1818,6 @@ fn function_has_body() {
     let indexed_crate = IndexedCrate::new(&crate_);
     let adapter = Arc::new(RustdocAdapter::new(&indexed_crate, None));
 
-    // Part 1: make sure unions have correct visibility (similart to importable_paths
-
     let query = r#"
 {
     Crate {
--- a/src/adapter/tests.rs
+++ b/src/adapter/tests.rs
@@ -1806,3 +1806,78 @@ fn unions() {
 
     similar_asserts::assert_eq!(expected_results, results);
 }
+
+#[test]
+fn function_has_body() {
+    let path = "./localdata/test_data/function_has_body/rustdoc.json";
+    let content = std::fs::read_to_string(path)
+        .with_context(|| format!("Could not load {path} file, did you forget to run ./scripts/regenerate_test_rustdocs.sh ?"))
+        .expect("failed to load rustdoc");
+
+    let crate_ = serde_json::from_str(&content).expect("failed to parse rustdoc");
+    let indexed_crate = IndexedCrate::new(&crate_);
+    let adapter = Arc::new(RustdocAdapter::new(&indexed_crate, None));
+
+    // Part 1: make sure unions have correct visibility (similart to importable_paths
+
+    let query = r#"
+{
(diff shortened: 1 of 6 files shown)


obi1kenobi:
Very nice, excellent work!

Would you mind adding some test cases? Check out `src/adapter/tests.rs` for the existing tests. You can create a new test crate with `cargo new <crate_name> --lib` and then add test code in it. Cases we probably want to cover:
- top-level function
- method inside an impl block
- trait method with / without body
- function in extern block: https://doc.rust-lang.org/std/keyword.extern.html

After we merge this, I'll take care of porting this PR to all the other rustdoc version numbers we support — I have some automation set up for that already.

obi1kenobi:
As an aside, I notice we are lacking a `CONTRIBUTING.md` file in this repo. Would you care to add one in a separate PR? No worries if that doesn't sound interesting, just offering more places where contributions are welcome :)

obi1kenobi:
I might have caught this before it was ready for review, sorry if that's the case.

I just noticed there are a couple of minor bugs in the test cases, and what looks like an accidental copy-paste artifact in the test code. You were probably already on the way to fixing those, I just wanted to mention them to be sure!

mrnossiom:
I consider the PR ready for review. I just don't see which minor bugs you are talking about.

test_crates/function_has_body/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+pub struct Foo;
+
+impl Foo {
+    pub fn inside_impl_block() {}
+}
+
+pub trait Bar {
+    fn trait_no_body();
+
+    fn trait_with_body();
```
obi1kenobi: This function doesn't actually have a body.
```suggestion
    fn trait_with_body() {}
```

src/adapter/tests.rs
```diff
@@ -1806,3 +1806,76 @@ fn unions() {
...
+            name: "inside_impl_block".into(),
+            has_body: true,
+        },
+        Output {
+            name: "trait_no_body".into(),
+            has_body: false,
+        },
+        Output {
+            name: "trait_with_body".into(),
+            has_body: false,
```
obi1kenobi: Given that we expect this function to have a body, `has_body` should be `true`:
```suggestion
            has_body: true,
```

obi1kenobi (approved):
Assuming my corrections are accurate, this should be good to merge!

I'll port it to the other rustdoc version branches, and I'll queue up releases for each version so you can use it in `cargo-semver-checks` by running `cargo update`.

mrnossiom:
🤦🏻 Thanks for fixing

obi1kenobi:
It's all good! Thanks for the PR!

I think you might have uncovered a bug in some of my code in the process, actually. It doesn't cause any correctness issues in cargo-semver-checks, but it might be negatively affecting performance. I'll have to look into it when I have a few minutes.

obi1kenobi:
#400 is the bug, in case you're curious. Thanks for helping surface it!

PR #402, opened 2024-08-21
Rayon feature

### LINKED ISSUE #401: Offer `rayon` crate feature that parallelizes index-building
Our `IndexedCrate` type builds a variety of indexes based on the underlying rustdoc crate.

Rustdoc crates' JSON can be ~450MB in size for particularly large crates, so our indexes can be significant in size too. They could probably be built in parallel using `rayon`.

We don't want to _require_ `rayon`, since that may not be suitable for all use cases. But we should offer it as a non-default option in case the performance improvement is worth it. For example, `cargo-semver-checks` would enable it, whereas the [rustdoc Trustfall playground](https://play.predr.ag/rustdoc) would probably not enable it.

Required steps:
- [ ] add optional `rayon` feature
- [ ] adjust CI to run tests both with and without the feature, so both configurations are known correct
- [ ] implement `rayon` parallelization for index-building
- [ ] benchmark (on your own machine) how much faster `rayon` makes `IndexedCrate::new()` on the rustdoc JSON of a large crate like [`aws-sdk-ec2`](https://github.com/awslabs/aws-sdk-rust/tree/main/sdk/ec2)
  - to generate rustdoc JSON, clone that repo, `cd` into the crate's directory, and use the `RUSTDOCFLAGS="-Z unstable-options --document-private-items --document-hidden-items --output-format=json --cap-lints=allow" cargo doc --lib --no-deps` shell command which will put `aws-sdk-ec2.json` into the `target/doc` directory of the workspace

This is a good first issue for someone who is already familiar with Rust, somewhat familiar with `rayon`, and wants to start contributing to this project. It is *not* a good first issue for folks new to Rust, since it requires understanding advanced Rust concepts such as `Send + Sync` auto-traits etc.

Comment to temporarily claim the issue, in case multiple people are interested in it!

jalil-salame:
A preliminary implementation of the `rayon` feature to speed up the
building of the `IndexedCrate` struct.

Results working on JSON from `aws-sdk-rust/aws-sdk-ec2`:

**Before:**

```
indexed_crate  fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ new         1.558 s       │ 1.675 s       │ 1.633 s       │ 1.631 s       │ 100     │ 100
```

**This commit:**

```
indexed_crate  fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ new         964.6 ms      │ 1.039 s       │ 997.9 ms      │ 998.7 ms      │ 100     │ 100
```

Closes #401

-----

I won't have time to work on this until this Friday (23rd August). So some thoughts in case someone wants to pick it up from here:

- Most of the gains come from parallelizing the `import_index`
- Parallelizing the `impl_index` properly is hard
- Some things benefit from being stored in a `BTreeMap` instead of a `HashMap` (or maybe pulling in a faster `Hash` impl)
- `.entry(key).or_default().push(value)` does not optimize as well as matching on the entry and inserting a `vec![value]` if vacant
- The `MapList` struct I introduced seems to hinder optimizations, so remove it where possible.
- Try to share more code between `rayon` and the default impl
- Switch the recursive functions for iterative ones as it shows up in the flamegraph

### DIFF
--- a/src/visibility_tracker.rs
+++ b/src/visibility_tracker.rs
@@ -2,6 +2,9 @@ use std::collections::{HashMap, HashSet};
 
 use rustdoc_types::{Crate, GenericArgs, Id, Item, ItemEnum, TypeAlias, Visibility};
 
+#[cfg(feature = "rayon")]
+use rayon::prelude::*;
+
 use crate::{attributes::Attribute, ImportablePath};
 
 #[derive(Debug, Clone)]
@@ -17,9 +20,14 @@ impl<'a> VisibilityTracker<'a> {
     pub(crate) fn from_crate(crate_: &'a Crate) -> Self {
         let mut visible_parent_ids = compute_parent_ids_for_public_items(crate_);
 
+        #[cfg(feature = "rayon")]
+        let iter = visible_parent_ids.par_iter_mut();
+        #[cfg(not(feature = "rayon"))]
+        let iter = visible_parent_ids.iter_mut();
+
         // Sort and deduplicate parent ids.
         // This ensures a consistent order, since queries can observe this order directly.
-        visible_parent_ids.iter_mut().for_each(|(_id, parent_ids)| {
+        iter.for_each(|(_id, parent_ids)| {
             parent_ids.sort_unstable_by_key(|id| id.0.as_str());
             parent_ids.dedup_by_key(|id| id.0.as_str());
         });


obi1kenobi:
Good stuff so far! Excited to see where you take this next 🚀

All perf gains you make here get doubled in `cargo-semver-checks`, since it loads two crate versions' data per run. So you've already cut over a second from the `cargo-semver-checks` of that crate!

When you're happy with the perf gains and are getting ready to merge this, it would be good to move the conditionally-compiled bits into separate functions that we can conditionally include and call. I think this might make the code a bit easier to maintain in the long run.

jalil-salame:
I tidied up the code and made sure it builds without rayon, I also profiled and most of the remaining time is spent on `VisibilityTracker::from_crate(crate_)` (roughly 60%-70% of the function time).

I'll refactor the `VisibilityTracker::from_crate(crate_)` function on Friday and see what perf gains I can achieve.

In the mean time could you check if the rayon stuff is easier to maintain now? The rayon and normal code is mixed together so that turning off the rayon feature affects the least amount of code.

benches/indexed_crate.rs
```diff
@@ -0,0 +1,14 @@
+use divan::Bencher;
+use trustfall_rustdoc_adapter::IndexedCrate;
+
+fn main() {
+    divan::main()
+}
+
+#[divan::bench]
+fn new(b: Bencher) {
+    let data = std::fs::read_to_string("/home/jalil/Dev/aws-sdk-rust/target/doc/aws_sdk_ec2.json")
```
obi1kenobi: I have a few concerns and a few suggestions on how to address them.

At a high level, the next person looking at this code in 6 months will probably not understand most of the context here. They won't know why this file was chosen, why that was a sound decision (as opposed to choosing any other file), how to create it, and almost certainly won't have a `/home/jalil/Dev` directory to place it in.

It would be good to automate the bits that can be automated, and thoroughly document the rest.

A better setup here would be:
- The benchmarking function runs a series of commands that, starting with "a machine with an internet connection," result in this file being available in a temp directory. If the file was already available (e.g. from a previous run), then it is reused as-is.
- The module uses a module-level doc comment (`//!` prefix starting on line 1) to describe why this crate's rustdoc JSON in particular is chosen: because it is very large, and highly representative of the perf environment facing the largest users of `trustfall-rustdoc-adapter`.

An example of the automated flow of commands could be:
- Check if the system's temp directory (whatever the OS!) contains a subdirectory called `aws-sdk-rust`. If it doesn't, `git clone` the relevant repo into such a subdirectory.
- Check out a specified commit hash (e.g. the current `HEAD` of the main branch) in that directory.
- Run a specified cargo command in that directory, causing the expected rustdoc JSON file to be created.
- If the specified commit hash was already checked out, and the expected rustdoc JSON file appeared to already be present at the expected location, skip the `git checkout` and `cargo` invocations and assume the file is good to reuse. *Document this behavior* in the module-level comment.
- Then, begin the benchmark using that file.

src/indexed_crate.rs
```diff
@@ -41,6 +41,218 @@ pub struct IndexedCrate<'a> {
     pub(crate) manually_inlined_builtin_traits: HashMap<Id, Item>,
 }
 
+#[cfg(feature = "rayon")]
+use rayon::prelude::*;
```
obi1kenobi: nit: it would be a bit better to move this adjacent to all the other imports, above `struct IndexedCrate`

src/indexed_crate.rs
```diff
@@ -41,6 +41,218 @@ pub struct IndexedCrate<'a> {
...
+        I: IntoParallelIterator<Item = (K, V)>,
+    {
+        par_iter
+            .into_par_iter()
+            .fold(Self::new, |mut map, (key, value)| {
+                map.insert(key, value);
+                map
+            })
+            .reduce(Self::new, |l, mut r| {
+                r.merge(l);
```
obi1kenobi: Interesting that we're merging to the right here. The reasoning behind this is probably worth explaining with a code comment.

src/indexed_crate.rs
```diff
@@ -41,6 +41,218 @@ pub struct IndexedCrate<'a> {
...
+            iter.map(move |importable_path| {
+                (importable_path.path, (item, importable_path.modifiers))
+            }),
+        )
+    })
+    .flatten()
+    .collect()
+}
+
+fn build_impl_index(index: &HashMap<Id, Item>) -> MapList<ImplEntry<'_>, (&Item, &Item)> {
```
obi1kenobi: Both of these two methods could use some doc comments that describe how they are different with and without the `rayon` feature. Those doc comments will help us create the documentation for the `rayon` feature itself, which will in turn help downstream maintainers decide whether that feature is worth enabling in their own use case.

src/indexed_crate.rs
```diff
@@ -41,6 +41,218 @@ pub struct IndexedCrate<'a> {
...
+        }
+    }
+
+    #[inline]
+    #[cfg(feature = "rayon")]
+    pub fn merge(&mut self, other: Self) {
+        for (key, value) in other.0 {
+            self.insert_many(key, value);
+        }
+    }
```
obi1kenobi: I'm wondering if the repeated merging of many `Vec<V>` together might result in a lot of copying that might be expensive.

Did you get a chance to experiment with a setup where the `merge()` works by appending `Vec<V>` into a `Vec<Vec<V>>`, and then when everything is fully collected, flattening it out exactly once into a `Vec<V>` that can be preallocated to exactly the right capacity?

Cargo.toml
```diff
@@ -13,9 +13,22 @@ readme = "./README.md"
...
+[features]
+default = ["rayon"]
+rayon = ["dep:rayon"]
+
+[[bench]]
+name = "indexed_crate"
+harness = false
+
+[profile.bench]
+debug = true
```
obi1kenobi: This feels like a very aggressive debug setting to enable by default in `bench` runs:
https://doc.rust-lang.org/cargo/reference/profiles.html#debug

Is this considered good practice for `bench` targets in other Rust projects? Would a less intensive version be sufficient for our perf-testing needs?

Cargo.toml
```diff
@@ -13,9 +13,22 @@ readme = "./README.md"
 [dependencies]
 trustfall = "0.7.1"
 rustdoc-types = "0.29.1"
+rayon = { version = "1.10.0", optional = true, default-features = false }
+
+[features]
+default = ["rayon"]
+rayon = ["dep:rayon"]
```
obi1kenobi: We'd like a fair bit of documentation to be available here, in order to help downstream maintainers decide if this feature is right for their use case.

We specifically want to describe what behavior changes this feature produces, and what it currently might not do but might start doing in the future.

obi1kenobi (commented):
The code is in reasonable shape, and is trending in an excellent direction. I'm excited to be able to merge it soon!

I've made some suggestions about how to make the code more maintainable in the long run, and more friendly to both "us in 6 months" and also to downstream maintainers who now need to decide if this new `rayon` feature is right for their use case.

Thanks for all your hard work so far! I'm excited to ship this soon 🚀

obi1kenobi:
Oh I almost forgot: we really want a CI step in the "run tests" job that ensures all tests pass both with and without the `rayon` feature.

src/indexed_crate.rs
jalil-salame: No, but I don't think the impact is big as I cannot see any `Vec` allocation related methods in the flamegraph.

Cargo.toml
jalil-salame: `cargo-flamegraph` is very pedantic about this so I added it to shut it up, I'll try with different values though!

Cargo.toml
jalil-salame: Will do! (also, the default to `rayon` is temporary for ease of development)

jalil-salame:
> Oh I almost forgot: we really want a CI step in the "run tests" job that ensures all tests pass both with and without the `rayon` feature.

Will do!

I wanted to get a Proof of Concept out ASAP so that I could ensure the things I'm doing align with your expectations/needs (and also to have something people could pick up and work with if I have to leave stuff halfway done).

I should have more time over the weekend to polish it up and get everything ready to merge c:

Cargo.toml
obi1kenobi: Ah TIL, that's also useful context to include. For example, `cargo-flamegraph` may change its recommendation in the future, and we want to know that this value is intended to match what `cargo-flamegraph` recommends and not due to some other thing. That will help us change the value appropriately in the future, without concerns about breaking something we don't understand!

benches/indexed_crate.rs
jalil-salame: Do we want to keep the benchmarks? I added them in a separate commit thinking I would drop it before marking it ready to merge (that's why they are in that state).

If we want them, I'll spend more time polishing them!

obi1kenobi:
Sounds great, and thank you for being super thoughtful about everything! The PR is definitely aligned with what I was hoping for, and I'm thrilled that you are approaching it with all that in mind! 🚀

benches/indexed_crate.rs
obi1kenobi: If you're open to polishing to the point where we merge them, that'd be sweet! Having nice, maintainable benchmarks will make it easier to merge future performance improvements both here and in `cargo-semver-checks`. We'll be able to make more ambitious changes on the basis of what we measure, as opposed to just going on intuition.

But if you don't want to do that work, I'd merge the PR without them too.

src/indexed_crate.rs
jalil-salame: Turns out merge left is faster, my intuition was wrong on this one XD, left a comment so it isn't swapped without measuring

jalil-salame:
I optimized the `VisibilityTracker::from_crate`. This shaves ~200ms off the numbers (no rayon 1.6s -> 1.4s) and with rayon (1s -> 0.8s).

There are probably some opportunities to parallelize it, but I'll work on that once I've had some sleep.

The main things I did were:
1. Turn the recursive function into an iterative function (drop interferes with tail calls so recursive functions are almost always a problem).
2. Prevent the rebuilding of the main `HashMap`: There was a `HashMap<&Id, HashSet<&Id>>` that we turned into a `HashMap<&Id, Vec<&Id>>`. This causes all keys to rehash, so if instead of using a `HashSet` we use a `Vec` directly and `sort+dedup` (we were already sorting the `Vec` so we don't pay that much extra time), we prevent the rehashing of the keys by doing an `iter_mut` instead of an `into_iter().collect()`

Cargo.toml
jalil-salame: AFAIK `cargo-flamegraph` is being too aggressive with its request, `debug = 1` seems to provide the same amount of information.

src/indexed_crate.rs
jalil-salame: Fixed!

src/indexed_crate.rs
jalil-salame: Fixed? Doc comments added, maybe they need more context though?

obi1kenobi:
Nice find! 200ms before `rayon` is great! Thank you for all your hard work 🚀

Since the changes to `VisibilityTracker::from_crate` aren't related to `rayon`, I'd find them a lot easier to review if you made them in a separate PR. Right now, they are quite difficult to analyze in a PR that does everything all at once.

That PR would also benefit from having a robust benchmark that's part of the repo already.

In an ideal world, we'd have separate PRs like this:
- benchmark we use to analyze performance
- serial improvements to `VisibilityTracker::from_crate`
- `rayon` improvements to `VisibilityTracker::from_crate`
- `rayon` improvements to `IndexedCrate`

The fewer lines and less complexity there is in each PR, the easier the code will be to understand and make maintainable. In turn, that means we'll probably be able to find even more performance wins in the future!

jalil-salame:
> Since the changes to `VisibilityTracker::from_crate` aren't related to `rayon`, I'd find them a lot easier to review if you made them in a separate PR. Right now, they are quite difficult to analyze in a PR that does everything all at once.

I don't know if you have seen this, but you can review commits individually: on the top left (right above the file tree in the review UI) you can change from `Changes from all commits` to changes from a specific commit. This adds a button to go to previous/next commit.

I can also split it up into multiple PRs, but because they all soft depend on each other I think this would be slightly easier to work on.

In preparation for either approach I split it up into multiple commits c:

benches/indexed_crate.rs
jalil-salame: I wrote my thoughts on the commit introducing benchmarks, but they are polished enough to merge rn

jalil-salame:
The profiler indicates that about ~80% of the time is now spent on `VisibilityTracker::from_crate`, and most of that time is spent on `HashMap` operations. Switching from `std::collections::HashMap` to `hashbrown::HashMap` or `rustc-hash::FxHashMap` might further improve performance.

An initial investigation on the improvements from switching to `rustc-hash::FxHashMap/Set` shows:

```console
$ cargo criterion --bench indexed_crate
IndexedCrate/new(aws-sdk-ec2)
                        time:   [1.3016 s 1.3060 s 1.3106 s]

$ cargo criterion --bench indexed_crate --features rustc-hash
IndexedCrate/new(aws-sdk-ec2)
                        time:   [1.1319 s 1.1352 s 1.1384 s]
                        change: [-13.469% -13.075% -12.695%] (p = 0.00 < 0.05)
                        Performance has improved.

$ cargo criterion --bench indexed_crate --features rayon
IndexedCrate/new(aws-sdk-ec2)
                        time:   [542.39 ms 543.88 ms 545.39 ms]
                        change: [-52.276% -52.090% -51.895%] (p = 0.00 < 0.05)
                        Performance has improved.

$ cargo criterion --bench indexed_crate --features rayon,rustc-hash
IndexedCrate/new(aws-sdk-ec2)
                        time:   [414.43 ms 415.55 ms 416.69 ms]
                        change: [-23.884% -23.596% -23.292%] (p = 0.00 < 0.05)
                        Performance has improved.
```

obi1kenobi:
> I don't know if you have seen this, but you can review commits individually: on the top left (right above the file tree in the review UI) you can change from Changes from all commits to changes from a specific commit. This adds a button to go to previous/next commit.
>
> I can also split it up into multiple PRs, but because they all soft depend on each other I think this would be slightly easier to work on.

I have seen it, yes. Unfortunately, individual commits have other downsides relative to separate PRs.

The core issue is that I maintain (currently) 5 different branches of this project at the same time. To make that feasible and not a nightmare, I have a workflow tuned around squash-merging PRs and cherry-picking the resulting PR commit onto all the different branches.

We have to keep the different commits here separate, in case in the future we find one of them has an issue that requires it to be reverted. That means I can't squash-merge, which means I can't use my usual automation for maintaining this repo.

Sorry! I hope you understand now why I have to insist on separate PRs.

jalil-salame:
> Sorry! I hope you understand now why I have to insist on separate PRs.

I understand, I assumed something like that would be the case, but I hoped it wouldn't be. Then I'll make PRs for each commit? Or should I group some of them together?

obi1kenobi:
PRs for each commit would be great! Feel free to open the independent ones concurrently — or all of them concurrently, if you don't mind some more rebasing as we merge them.

Thanks for bearing with me here, I know this is an inconvenience and appreciate you being open to helping make my workflow easier!

jalil-salame:
I split off the first three commits as they can be merged in parallel, the rest should cause merge conflicts so I'll create the PRs once they are merged.

My system seems to be noisy for perf measurements so expect results to be off by +/-2% (even when criterion says something is 2% faster, it's most likely noise).

jalil-salame:
Marked as draft to prevent accidental merges c:

jalil-salame:
All parts of this merge request have been  included!

PR #403, opened 2024-08-26
feat: add benchmarks for the IndexedCrate::new method

jalil-salame:
Currently uses `criterion` for benchmarks, but another option would be `iai-callgrind` which provides "oneshot" benchmarks (which are way faster and more reliable in busy systems like CI), but AFAIK `iai-callgrind` only supports Linux.

Also adds a script to download aws-sdk-ec2 and build the rustdoc JSON. It should work on MacOS, but I don't have a way to test that. It also relies on `rustup` (because we use `cargo +nightly`).

Current numbers on my system:

```
IndexedCrate/new(aws-sdk-ec2)
                        time:   [1.5561 s 1.5609 s 1.5656 s]
```

Split off from #402

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,8 @@ test_crates/**/Cargo.lock
 
 # ide
 .idea
+
+# Benchmarks and profiling related data
+flamegraph.svg
+perf.data
+perf.data.old
--- a/scripts/prepare_benchmark_data.sh
+++ b/scripts/prepare_benchmark_data.sh
@@ -1,0 +1,54 @@
+#!/bin/sh
+
+aws_sdk_version="${AWS_SDK_VERSION:-release-2024-08-22}"
+
+bail() {
+	echo "ERROR: $1"
+	exit 1
+}
+
+repo_path="$(realpath "$(dirname "$0")")"/..
+bench_data_path="${repo_path}/localdata/benches/aws-sdk-ec2.json"
+
+mkdir -p "${repo_path}/localdata/benches"
+
+if [ -f "$bench_data_path" ]; then
+	echo "Reusing bench input data: already present at $bench_data_path"
+	echo "help: delete the file if you want to regenerate it"
+	echo "      rm -v '$bench_data_path'"
+	exit 0
+fi
+
+# FIXME: This is a Linux specific path, on MacOS it should instead point to $HOME/Library/Caches/
+#        see: https://docs.rs/directories/latest/directories/struct.ProjectDirs.html#method.cache_dir
+cache_path="${XDG_CACHE_HOME:-$HOME/.cache}"
+download_url="https://github.com/awslabs/aws-sdk-rust/archive/refs/tags/${aws_sdk_version}.tar.gz"
+
+source_tarball="${cache_path}/aws-sdk-rust-${aws_sdk_version}.tar.gz"
+
+if [ -f "$source_tarball" ]; then
+	echo "Reusing source tarball: already present at $source_tarball"
+	echo "help: delete the file if you want to download it again"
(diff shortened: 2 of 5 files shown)


scripts/prepare_benchmark_data.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# FIXME: This is a Linux specific path, on MacOS it should instead point to $HOME/Library/Caches/
+#        see: https://docs.rs/directories/latest/directories/struct.ProjectDirs.html#method.cache_dir
+cache_path="${XDG_CACHE_HOME:-$HOME/.cache}"
+download_url="https://github.com/awslabs/aws-sdk-rust/archive/refs/tags/${aws_sdk_version}.tar.gz"
+
+source_tarball="${cache_path}/aws-sdk-rust-${aws_sdk_version}.tar.gz"
+
+if [ -f "$source_tarball" ]; then
+	echo "Source tarball already present at: $source_tarball"
```
obi1kenobi: A small linguistic nudge, just to indicate this isn't an error condition:
```suggestion
	echo "Reusing source tarball already present at: $source_tarball"
```

scripts/prepare_benchmark_data.sh
```diff
@@ -0,0 +1,52 @@
...
+	echo "      rm -v '$source_tarball'"
+else
+	echo "Downloading sources..."
+	curl --proto '=https' --tlsv1.2 -SfL "$download_url" -o "$source_tarball" || bail "failed to download aws-sdk-rust sources"
+fi
+
+source_dir="${cache_path}/aws-sdk-rust-${aws_sdk_version}"
+
+if [ -d "$source_dir" ]; then
+	echo "Source already present at: $source_dir"
```
obi1kenobi: Similar to the above:
```suggestion
	echo "Reusing source already present at: $source_dir"
```

scripts/prepare_benchmark_data.sh
```diff
@@ -0,0 +1,52 @@
...
+	echo "help: delete the file if you want to extract it again"
```
obi1kenobi: Presumably a directory needs to be deleted, right?
```suggestion
	echo "help: delete the directory if you want to extract the contents again"
```

.gitignore
```diff
@@ -16,3 +16,9 @@ test_crates/**/Cargo.lock
 
 # ide
 .idea
+
+# bench
+flamegraph.svg
+perf.data
+perf.data.old
+benches/aws-sdk-ec2.json
```
obi1kenobi: Would it be possible to place this inside e.g. `localdata/benches/` instead?

`localdata/` is where the tests' rustdoc JSONs are stored, so the symmetry would be nice there.

obi1kenobi (commented):
Broadly, looks great! Just a couple of tiny suggestions and nitpicks for your consideration.

scripts/prepare_benchmark_data.sh
jalil-salame: Fixed!

scripts/prepare_benchmark_data.sh
jalil-salame: Fixed!

scripts/prepare_benchmark_data.sh
jalil-salame: Fixed!

.gitignore
jalil-salame: `flamegraph.svg` and `perf.data*` are dumped in the CWD, but I moved the `aws-sdk-ec2.json`

obi1kenobi (approved):

PR #404, opened 2024-08-26
feat: faster VisibilityTracker::from_crate

jalil-salame:
We reduce the number of times `&Id`s get hashed and get a nice perf boost thanks to that.

Numbers on my machine:

```
IndexedCrate/new(aws-sdk-ec2)
                        time:   [1.3597 s 1.3602 s 1.3607 s]
                        change: [-13.397% -12.945% -12.503%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Split off from #402 (perf numbers come from the commits there)

### DIFF
--- a/src/visibility_tracker.rs
+++ b/src/visibility_tracker.rs
@@ -15,15 +15,14 @@ pub(crate) struct VisibilityTracker<'a> {
 
 impl<'a> VisibilityTracker<'a> {
     pub(crate) fn from_crate(crate_: &'a Crate) -> Self {
-        let visible_parent_ids = compute_parent_ids_for_public_items(crate_)
-            .into_iter()
-            .map(|(key, values)| {
-                // Ensure a consistent order, since queries can observe this order directly.
-                let mut values: Vec<_> = values.into_iter().collect();
-                values.sort_unstable_by_key(|x| &x.0);
-                (key, values)
-            })
-            .collect();
+        let mut visible_parent_ids = compute_parent_ids_for_public_items(crate_);
+
+        // Sort and deduplicate parent ids
+        // This ensures a consistent order, since queries can observe this order directly.
+        visible_parent_ids.iter_mut().for_each(|(_id, parent_ids)| {
+            parent_ids.sort_unstable_by_key(|id| id.0.as_str());
+            parent_ids.dedup_by_key(|id| id.0.as_str());
+        });
 
         Self {
             inner: crate_,
@@ -249,8 +248,7 @@ struct NameResolution<'a> {
     duplicated_glob_names_in_module: HashMap<&'a Id, HashSet<NamespacedName<'a>>>,
 }
 
-fn compute_parent_ids_for_public_items(crate_: &Crate) -> HashMap<&Id, HashSet<&Id>> {
-    let mut result = Default::default();
+fn compute_parent_ids_for_public_items(crate_: &Crate) -> HashMap<&Id, Vec<&Id>> {
     let root_id = &crate_.root;
 
     if let Some(root_module) = crate_.index.get(root_id) {
@@ -260,6 +258,8 @@ fn compute_parent_ids_for_public_items(crate_: &Crate) -> HashMap<&Id, HashSet<&
             // Avoid cycles by keeping track of which items we're in the middle of visiting.
             let mut currently_visited_items: HashSet<&Id> = Default::default();
 
+            let mut result = Default::default();
+
             visit_root_reachable_public_items(
                 crate_,
                 &mut result,
@@ -268,10 +268,12 @@ fn compute_parent_ids_for_public_items(crate_: &Crate) -> HashMap<&Id, HashSet<&
                 root_module,
                 None,
             );
+
+            return result;
         }
     }
 
-    result
+    Default::default()
 }
 
 fn get_names_for_item<'a>(
@@ -471,7 +473,7 @@ fn resolve_glob_imported_names<'a>(crate_: &'a Crate, traversal_state: &mut Name
 
         // Glob-of-glob import chains might still produce `names` and `duplicated_names` entries
         // that would be shadowed by locally-defined names in this module. Apply the shadowing
-        // rules by removing any conflicing names from both of those collections.
+        // rules by removing any conflicting names from both of those collections.
         if let Some(local_names) = traversal_state.names_defined_in_module.get(module_id) {
             for local_name in local_names.keys() {
                 names.remove(local_name);
@@ -606,7 +608,7 @@ fn register_name<'a>(
 /// Collect all public items that are reachable from the crate root and record their parent Ids.
 fn visit_root_reachable_public_items<'a>(
     crate_: &'a Crate,
-    parents: &mut HashMap<&'a Id, HashSet<&'a Id>>,
+    parents: &mut HashMap<&'a Id, Vec<&'a Id>>,
     traversal_state: &NameResolution<'a>,
     currently_visited_items: &mut HashSet<&'a Id>,
     item: &'a Item,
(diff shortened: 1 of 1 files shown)


src/visibility_tracker.rs
```diff
@@ -15,15 +15,15 @@ pub(crate) struct VisibilityTracker<'a> {
 
 impl<'a> VisibilityTracker<'a> {
     pub(crate) fn from_crate(crate_: &'a Crate) -> Self {
-        let visible_parent_ids = compute_parent_ids_for_public_items(crate_)
-            .into_iter()
-            .map(|(key, values)| {
-                // Ensure a consistent order, since queries can observe this order directly.
```
obi1kenobi, line 21: Let's try to preserve this comment, since I believe it's still relevant and accurate. Adding to my suggestion in the other comment.

src/visibility_tracker.rs
```diff
@@ -15,15 +15,15 @@ pub(crate) struct VisibilityTracker<'a> {
...
-            .collect();
+        let mut visible_parent_ids = compute_parent_ids_for_public_items(crate_);
+
+        let iter = visible_parent_ids.iter_mut();
+
+        // Sort and deduplicate parent ids
+        iter.for_each(|(_id, parent_ids)| {
+            parent_ids.sort_unstable_by_key(|id| id.0.as_str());
+            parent_ids.dedup_by_key(|id| id.0.as_str());
+        });
```
obi1kenobi: A bit of personal preference, if you don't mind indulging me: I find it easier to understand iterator `.for_each()` when the base iterator is constructed inline instead of constructed separately and assigned to a variable.
```suggestion
        // Sort and deduplicate parent ids.
        // This ensures a consistent order, since queries can observe the order directly.
        visible_parent_ids.iter_mut().for_each(|(_id, parent_ids)| {
            parent_ids.sort_unstable_by_key(|id| id.0.as_str());
            parent_ids.dedup_by_key(|id| id.0.as_str());
        });
```

src/visibility_tracker.rs
```diff
@@ -606,198 +595,204 @@ fn register_name<'a>(
...
+        }
+
+        if !currently_visited_items.insert(item.id.0.as_str()) {
+            // We found a cycle in the import graph, and we've already processed this item.
+            // Nothing more to do here.
+            continue;
+        }
+
+        let next_parent_id = Some(&item.id);
+        // queue up removing this from the visited list
```
obi1kenobi: Just to match the style of the surrounding comments.
```suggestion
        // Queue up removing this from the visited list.
```

src/visibility_tracker.rs
```diff
@@ -606,198 +595,204 @@ fn register_name<'a>(
...
+        stack.push(Action::Remove { id: &item.id });
+
+        match &item.inner {
+            rustdoc_types::ItemEnum::Module(m) => {
+                stack.reserve(m.items.len());
+                stack.extend(m.items.iter().filter_map(|id| {
```
obi1kenobi: It's interesting that this does `.reserve()` then `.extend()`.

The iterator length of `m.items.iter().filter_map()` is known to be `<= m.items.len()`. Is this not something that the stack's `.extend()` will already account for?

src/visibility_tracker.rs
```diff
@@ -606,198 +595,204 @@ fn register_name<'a>(
...
-                    next_parent_id,
-                );
+            rustdoc_types::ItemEnum::Trait(trait_) => {
+                stack.reserve(trait_.items.len());
+                stack.extend(trait_.items.iter().filter_map(|id| {
+                    crate_.index.get(id).map(|inner| Action::Process {
+                        item: inner,
+                        parent_id: next_parent_id,
+                    })
+                }));
```
obi1kenobi: Do you think it'd be feasible to make some kind of refactor to deduplicate this `reserve() + extend() + crate_.index.get(id).map(Action::Process {...})` pattern?

This method was pretty complex to start with, and now includes even more complexity with a fair bit of extra repetition.

src/visibility_tracker.rs
```diff
@@ -606,198 +595,204 @@ fn register_name<'a>(
...
-            // reachable items with default visibility must be public.
-        }
-    }
-
-    let item_parents = parents.entry(&item.id).or_default();
-    if let Some(parent_id) = parent_id {
-        item_parents.insert(parent_id);
+) -> HashMap<&'a Id, Vec<&'a Id>> {
+    #[derive(Debug)]
+    enum Action<'a> {
```
obi1kenobi: It would be good to leave a doc comment explaining the purpose of this type.

It took me a non-trivial amount of time to figure out why it's present and how it works. I'm not confident that in 6 months you or I will still remember the details, so as a favor to our future selves, I'd like to err on the side of explaining in too much detail rather than too little.

src/visibility_tracker.rs
```diff
@@ -606,198 +595,204 @@ fn register_name<'a>(
...
-        return;
-    }
+    let mut parents: HashMap<&'a Id, Vec<&'a Id>> = Default::default();
+    // Avoid cycles by keeping track of which items we're in the middle of visiting.
+    let mut currently_visited_items: HashSet<&'a str> = Default::default();
+
+    let mut stack = vec![Action::Process {
+        item,
+        parent_id: None,
+    }];
```
obi1kenobi: Similarly here, it would be good to leave an explanatory comment for what's happening: we're simulating recursion with a stack, the different actions mean A and B and correspond to equivalent operations X and Y in the recursive model. The stack-based implementation is faster because of reasons XYZ etc.

All this will be useful context for the next person who might try to update or optimize this code further.

obi1kenobi (commented):
This is a great optimization, nice find!

I'd love to make the new code a bit more maintenance-friendly before we merge it. As Rust evolves, we'll need to keep updating this code to cover new item kinds and it'll be important that we always understand how the code works and why. Otherwise, we'll write bugs and performance regressions, which will be frustrating to everyone.

jalil-salame:
> This is a great optimization, nice find!
> 
> I'd love to make the new code a bit more maintenance-friendly before we merge it. As Rust evolves, we'll need to keep updating this code to cover new item kinds and it'll be important that we always understand how the code works and why. Otherwise, we'll write bugs and performance regressions, which will be frustrating to everyone.

Lets push this PR back a bit. I'll split it into the manual dedup and the rewrite, because the more i look at the rewrite, the less i like it.

obi1kenobi:
Sounds good to me :+1:

src/visibility_tracker.rs
jalil-salame: Fixed (by removing the refactor)

src/visibility_tracker.rs
jalil-salame: Fixed (by removing the refactor)

src/visibility_tracker.rs
jalil-salame: Fixed (by removing the refactor)

src/visibility_tracker.rs
jalil-salame: Fixed (by removing the refactor)

src/visibility_tracker.rs
jalil-salame: Fixed (by removing the refactor)

src/visibility_tracker.rs
jalil-salame: Fixed!

src/visibility_tracker.rs
jalil-salame: Fixed!

src/visibility_tracker.rs
```diff
@@ -15,15 +15,14 @@ pub(crate) struct VisibilityTracker<'a> {
...
-            .map(|(key, values)| {
-                // Ensure a consistent order, since queries can observe this order directly.
-                let mut values: Vec<_> = values.into_iter().collect();
-                values.sort_unstable_by_key(|x| &x.0);
-                (key, values)
-            })
-            .collect();
+        let mut visible_parent_ids = compute_parent_ids_for_public_items(crate_);
+
+        // Sort and deduplicate parent ids
```
obi1kenobi:
```suggestion
        // Sort and deduplicate parent ids.
```

obi1kenobi (approved):

PR #405, opened 2024-08-26
chore: refactor IndexedCrate::new in preparation for parallelizing

jalil-salame:
Modify imperative code into iterator code to make it easier to parallelize with rayon.

No perf increase:

```
IndexedCrate/new(aws-sdk-ec2)
                        time:   [1.3491 s 1.3505 s 1.3520 s]
                        change: [-0.8260% -0.7120% -0.6178%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

Split off from #402 (perf numbers come from the commits there)

### DIFF
--- a/src/indexed_crate.rs
+++ b/src/indexed_crate.rs
@@ -186,7 +186,7 @@ impl<'a> IndexedCrate<'a> {
         // Build the imports index
         //
         // This is inlined because we need access to `value`, but `value` is not a valid
-        // `IndexedCrate` yet, do not extract into a separate function.
+        // `IndexedCrate` yet. Do not extract into a separate function.
         value.imports_index = Some(
             crate_
                 .index
--- a/src/indexed_crate.rs
+++ b/src/indexed_crate.rs
@@ -1,6 +1,6 @@
 use std::{
     borrow::Borrow,
-    collections::{BTreeSet, HashMap},
+    collections::{hash_map::Entry, HashMap, HashSet},
 };
 
 use rustdoc_types::{Crate, Id, Item};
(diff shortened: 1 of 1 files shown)


src/indexed_crate.rs
```diff
@@ -41,6 +41,166 @@ pub struct IndexedCrate<'a> {
...
+struct MapList<K, V>(HashMap<K, Vec<V>>);
+
+impl<K: std::cmp::Eq + std::hash::Hash, V> FromIterator<(K, V)> for MapList<K, V> {
+    #[inline]
+    fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> Self {
+        let iter = iter.into_iter();
+        let (min, max) = iter.size_hint();
+        // Reserving this space doesn't give a measurable gain, but neither does it hurt so we are
+        // keeping it around
+        let mut map = Self::with_capacity(max.unwrap_or(min));
```
obi1kenobi: If this doesn't produce measurable gain, I'd prefer to not include it for complexity and maintenance reasons.

Given that reserving space sometimes seems to make performance worse, I think it's a risk with larger downside than upside. We don't have enough test infrastructure set up to monitor for and prevent performance regressions, so we should minimize the risk they happen unnoticed.

I like the comment below that says "we tried reserving space and this was the result." Would you mind switching to something similar here?

src/indexed_crate.rs
```diff
@@ -41,6 +41,166 @@ pub struct IndexedCrate<'a> {
...
+    }
+}
+
+/// Build the imports index
+///
+/// When compiled using the `rayon` feature, build it in parallel. Specifically, this paralelizes
+/// getting the [`IndexedCrate::publicly_importable_names`] over the items in the index.
+fn build_imports_index<'a>(
+    index: &'a HashMap<Id, Item>,
+    value: &IndexedCrate<'a>,
```
obi1kenobi: This `&IndexedCrate<'a>` is actually not fully constructed at the time of this call. That is a potential footgun that could waste a few hours in a future iteration of tweaking this code, and I'd prefer to avoid introducing it in the first place.

Would you mind inlining this method back into `IndexedCrate::new()` where it's much more clear that the `value` is not fully constructed yet?

It might even be good to add an explicit comment about how it's not fully constructed at that point of use, to explain why extracting that complexity into its own method isn't desirable. In retrospect, that isn't obvious and I totally understand why you moved it here as a result 😅

src/indexed_crate.rs
```diff
@@ -41,6 +41,166 @@ pub struct IndexedCrate<'a> {
...
+    .collect()
+}
+
+/// Build the impl index
+///
+/// When compiled using the `rayon` feature, build it in parallel. Specifically, this paralelizes
+/// the work of gathering all of the impls for the items in the index.
+fn build_impl_index(index: &HashMap<Id, Item>) -> MapList<ImplEntry<'_>, (&Item, &Item)> {
+    let iter = index.iter();
+    iter.filter_map(|(id, item)| {
```
obi1kenobi, line 102: I mentioned in a parallel PR that as a personal preference thing, I find code easier to read if it creates and immediately consumes iterators instead of going through an intermediate variable.

If you don't mind, I'd appreciate it if you could tweak the code here and elsewhere in this PR in that manner, to help me maintain this code with less work in the long run:
```suggestion
    index.iter().filter_map(|(id, item)| {
```

obi1kenobi (commented):
Thanks for putting this together and splitting up the PR, I really appreciate it!

I have a few minor comments, otherwise this is good to merge!

src/indexed_crate.rs
jalil-salame: We introduce the variable back once we parallelize it with rayon:

```rust
#[cfg(feature = "rayon")]
let iter = iter.par_iter();
#[cfg(not(feature = "rayon"))]
let iter = iter.iter();
```

I also prefer it the other way around.

I can switch it around as then reverting the rayon feature would leave the code in a better state.

obi1kenobi: Ah that explains it, thank you.

I agree, adding the variable in the `rayon`-specific PR would be ideal.

obi1kenobi:
Btw I just made a Discord server for cargo-semver-checks and Trustfall, to make coordination a bit easier. You're welcome to join if you'd like: https://discord.gg/s7Cfmp3P

src/indexed_crate.rs
jalil-salame: Fixed!

src/indexed_crate.rs
jalil-salame: Fixed!

src/indexed_crate.rs
jalil-salame: No problem!

src/indexed_crate.rs
```diff
@@ -51,95 +183,28 @@ impl<'a> IndexedCrate<'a> {
...
-                {
-                    if let Some(contained_item_name) = contained_item.name.as_deref() {
-                        impl_index
-                            .entry(ImplEntry::new(id, contained_item_name))
-                            .or_default()
-                            .push((impl_item, contained_item));
+        // Build the imports index
+        //
+        // This is inlined because we need access to `value`, but `value` is not a valid
+        // `IndexedCrate` yet, do not extract into a separate function.
```
```suggestion
        // `IndexedCrate` yet. Do not extract into a separate function.
```

obi1kenobi (approved):

PR #424, opened 2024-08-28
feat: parallelize VisibilityTracker::from_crate

### LINKED ISSUE #401: Offer `rayon` crate feature that parallelizes index-building

jalil-salame:
Small perf gains as there isn't much we can parallelize without more code massaging:

```
(92a9868) ❯ cargo criterion --bench indexed_crate --features rayon
IndexedCrate/new(aws-sdk-ec2)
                        time:   [574.13 ms 575.49 ms 576.92 ms]

(changes) ❯ cargo criterion --bench indexed_crate --features rayon
IndexedCrate/new(aws-sdk-ec2)
                        time:   [557.73 ms 559.38 ms 560.96 ms]
                        change: [-3.1709% -2.8003% -2.4380%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Last part of #402

### DIFF
--- a/src/visibility_tracker.rs
+++ b/src/visibility_tracker.rs
@@ -2,6 +2,9 @@ use std::collections::{HashMap, HashSet};
 
 use rustdoc_types::{Crate, GenericArgs, Id, Item, ItemEnum, TypeAlias, Visibility};
 
+#[cfg(feature = "rayon")]
+use rayon::prelude::*;
+
 use crate::{attributes::Attribute, ImportablePath};
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


obi1kenobi (approved):

obi1kenobi:
Resolves #401.

PR #429, opened 2024-08-30
After today, beta Rust uses rustdoc v33 so we should test on it.

obi1kenobi:


jalil-salame:
Seems like CI's beta is outdated (still uses `rustdoc v32`), you might want to rerun it

obi1kenobi:
I think the new beta wasn't out yet when I made this PR, I just wanted to
make sure to not forget :)

PR #435, opened 2024-08-31
feat: introduce internal IdRef type

jalil-salame:
We use `&Id` in many places. As `Id` is defined as `struct Id(pub String)` this is a ptr-to-ptr situation. By using `&IdRef(str)` we remove that indirection, this increases memory usage as `&IdRef` is now a "wide" ptr (ptr and length) instead of a "thin" ptr, but the perf win is worth it:

```console
(32418dd)❯ cargo criterion --bench indexed_crate
                        time:   [1.3590 s 1.3599 s 1.3609 s]

(changes)❯ cargo criterion --bench indexed_crate
IndexedCrate/new(aws-sdk-ec2)
                        time:   [1.2456 s 1.2466 s 1.2478 s]
                        change: [-8.4413% -8.3275% -8.2222%] (p = 0.00 < 0.05)
                        Performance has improved.

(32418dd)❯ cargo criterion --bench indexed_crate --features rayon
IndexedCrate/new(aws-sdk-ec2)
                        time:   [571.63 ms 572.93 ms 574.23 ms]

(changes)❯ cargo criterion --bench indexed_crate --features rayon
IndexedCrate/new(aws-sdk-ec2)
                        time:   [501.28 ms 502.32 ms 503.34 ms]
                        change: [-12.596% -12.324% -12.053%] (p = 0.00 < 0.05)
                        Performance has improved.
```

The discussions on Zulip propose changing the `Id` definition in `rustdoc-types` to `u64` or even `u32`, but while we wait for this change to land (and for older `rustdoc` versions), this should be a nice perf improvement.

The improvements are bigger when using `rayon`, because we spend more time on `visibility_tracker` stuff (which is where the biggest improvements are).

### DIFF
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -472,7 +472,7 @@ pub(super) fn resolve_trait_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
 
             let trait_vertex = vertex.as_trait().expect("not a Trait vertex");
             Box::new(trait_vertex.bounds.iter().filter_map(move |bound| {
-                if let TraitBound { trait_, .. } = bound {
+                if let TraitBound { trait_, .. } = &bound {
                     item_index
                         .get(&trait_.id)
                         .as_ref()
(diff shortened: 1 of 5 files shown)


obi1kenobi (approved):
Awesome job as always! 🚀

PR #508, opened 2024-10-15
Use elided lifetime parameters when possible.

obi1kenobi:


obi1kenobi:
Merged as part of #509 since they depended on each other to unbreak CI.

PR #528, opened 2024-10-16
Support for printing type name strings

suaviloquence:
This PR updates the `RawType.typename` property to serialize most of the types that `rustdoc_types` exposes into String.

I would like help figuring out how to test the many different cases.  The `rustdoc-adapter` seems to only expose `raw_type` on a struct field, and there are types (especially with generics) that I don't think are expressible as struct field types, but I don't really know how to wrangle the `rustdoc` JSON output to test e.g. function parameter types or associated types.


test_crates/typename/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub type B = ();
+}
+
+pub struct Struct<'a, T> {
+    pub a: String,
+    pub b: T,
+    pub c: Option<T>,
+    pub d: <A as MyTrait>::A<'static, ()>,
+    pub e: a::B,
+    pub f: unsafe extern "C-unwind" fn() -> T,
```
obi1kenobi: Function pointers can be unsafe too? TIL, better add more semver lints to the list... 😅

src/adapter/typename.rs
```diff
@@ -0,0 +1,448 @@
...
+            write!(fmt, "{delim}")?;
+        }
+
+        f(item, fmt)?;
+    }
+
+    Ok(())
+}
+
+#[inline]
```
obi1kenobi: We might not want this `#[inline]`. Did you get a chance to measure whether it helps, or was it just a momentary judgment call?

This is something I learned recently: inlining in LLVM happens from the leaves up. So it's quite likely that the `intersperse()` function will get inlined *into here*, and then this `#[inline]` might inline this call upward. That might not be desirable, and certainly isn't obvious.

src/adapter/mod.rs
```diff
@@ -21,6 +21,7 @@ mod enum_variant;
 mod optimizations;
 mod origin;
 mod properties;
+mod typename;
```
obi1kenobi: `typename` might be a confusing naming collision between Trustfall's concept of the `__typename` special property, implemented with trait `Typename` and its `typename()` method, and Rust type names.

Consider `rust_type_name` or some similar name?

src/adapter/typename.rs
```diff
@@ -0,0 +1,448 @@
...
+    items: I,
+    f: F,
+) -> Result {
+    intersperse(fmt, delim, items, |x, fmt| f(x).fmt(fmt))
+}
+
+display_wrapper! {
+    GenericParamDef,
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        match &self.0.kind {
```
obi1kenobi: Complex functions defined inside a macro can be quite annoying from a maintenance perspective. The experience of using editor integrations noticeably degrades when inside a macro, since tools like r-a don't know if there's special magic involved or not.

That said, I think it's totally reasonable to have a macro like this. So I'd propose what I believe is a best-of-both-worlds approach:
- Extract the complex logic into `#[inline]` functions with signatures `(value: &rustdoc_types::SomeType, fmt: &mut Formatter<'_>) -> Result`.
- Make the macro take an identifier for the formatter function to call, instead of taking an entire function item.
- Then have the macro generate the entire `fn fmt(&self, ...)` item itself, since it's always the same and just calls `specified_formatter(self.0, fmt)` under the hood.

src/adapter/typename.rs
```diff
@@ -0,0 +1,448 @@
...
+                Ok(())
+            }
+            rustdoc_types::Type::Tuple(tys) => {
+                write!(f, "(")?;
+                intersperse_with(f, ", ", tys, Type)?;
+                write!(f, ")")
+            }
+            rustdoc_types::Type::Slice(ty) => write!(f, "[{}]", Type(ty)),
+            // according to docs, `len` is not guaranteed to roundtrip
+            rustdoc_types::Type::Array { type_, len } => write!(f, "[{}; {}]", Type(type_), len),
```
obi1kenobi: Could you elaborate on this comment, and ideally provide a link to those docs?

It sounds kind of ominous maintenance-wise, as a "this might break at any time" sort of thing. It would be nice to make the _intent_ behind leaving this comment clear, and provide a location to learn more.

obi1kenobi (commented):
Excellent work, very impressively done! I love the attention to tests and the effort to make the code composable and "obviously correct."

I have a few minor suggestions for tweaks that should be resolvable quickly, and this will be ready to merge 🎉

obi1kenobi:
> I would like help figuring out how to test the many different cases. The `rustdoc-adapter` seems to only expose `raw_type` on a struct field, and there are types (especially with generics) that I don't think are expressible as struct field types, but I don't really know how to wrangle the `rustdoc` JSON output to test e.g. function parameter types or associated types.

The best way to more thoroughly test this code would be in isolation from Trustfall: load the JSON as `rustdoc-types`, look up the item you want to test, and test it directly.

I'd leave the existing Trustfall-based test in — it's a reasonable and cheap way to get more coverage. But in an ideal world, we'd _directly_ cover things like fn parameters and return types, associated types, `pub type` types, and other things that the adapter doesn't expose currently. That will help us be more confident in exposing them in the future, in some more stable and permanent manner than the perma-unstable `RawType` schema type.

src/adapter/typename.rs
suaviloquence: Hm, I didn't know that.  I just didn't want the overhead of a function call here, but LLVM may figure that out on its own, since we don't want to inline the `intersperse` call.

src/adapter/rust_type_name.rs
```diff
@@ -0,0 +1,464 @@
...
+                intersperse(f, " + ", outlives, String::fmt)?;
+            }
+        }
+        rustdoc_types::GenericParamDefKind::Type {
+            bounds,
+            default,
+            is_synthetic,
+        } => {
+            if *is_synthetic {
+                todo!("is_synthetic generic");
```
obi1kenobi: If you'd like a test case for this, write a function that uses `impl Trait` in argument position.

In rustdoc JSON, you'll get a synthetic generic named `impl Trait` that corresponds to the unnamed generic type of that argument.

src/adapter/rust_type_name.rs
```diff
@@ -0,0 +1,464 @@
...
+
+            Ok(())
+        }
+        rustdoc_types::Type::Tuple(tys) => {
+            write!(f, "(")?;
+            intersperse_with(f, ", ", tys, Type)?;
+            write!(f, ")")
+        }
+        rustdoc_types::Type::Slice(ty) => write!(f, "[{}]", Type(ty)),
+        // according to docs, `len` is not guaranteed to roundtrip
```
obi1kenobi: I think I had a comment on this line in a previous review that has disappeared due to the refactor here.

src/adapter/rust_type_name.rs
```diff
@@ -0,0 +1,464 @@
...
+            match modifier {
+                rustdoc_types::TraitBoundModifier::Maybe => write!(f, "?")?,
+                // TODO: check this/find a good reference.  it's currently unstablke
+                rustdoc_types::TraitBoundModifier::MaybeConst => write!(f, "~const ")?,
+                rustdoc_types::TraitBoundModifier::None => (),
+            };
+
+            write!(f, "{}", Path(trait_))?;
+
+            if generic_params.is_empty() {
```
obi1kenobi: Should this be a `!is_empty()` instead?

Consider writing a test case to trigger this bug before fixing it, since it's a sneaky one.

src/adapter/rust_type_name.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+}
+
+display_wrapper!(GenericBound, fmt_generic_bound);
+
+fn fmt_constant(this: &Constant, f: &mut Formatter<'_>) -> Result {
+    if let Some(val) = &this.0.value {
+        write!(f, "{}", val)
+    } else {
+        // TODO: investigate expr docs - stringified is unstable
```
obi1kenobi: Consider elaborating on this TODO note. Ideally, we'd like to make it clear what the action item is and why it's important, in case we decide it isn't worth addressing immediately and then someone else down the line inherits this code.

obi1kenobi (commented):
The refactor looks good! I think I caught a minor bug, and would love to get a touch more test coverage and docs in a few places before we merge. But nothing major!

src/adapter/rust_type_name.rs
suaviloquence: I added a test case, but it's not triggering the `todo!`, so `is_synthetic` is not true here? I'm not sure.

src/adapter/rust_type_name.rs
suaviloquence: Fixed, this will be caught by e.g. `impl for<'a> Fn(&'a ())` which is tested.

suaviloquence:
For test thoroughness, I wrote a simple test for most things in the top-level `Type` enum, as well as the function `my_generic_function` which tests some complicated types.  What other tests should I add?

I also need to fix `&'_ {dyn,impl} A + B + ...`, which needs to be `&'_ (dyn ...)`.  I think this is the only case where we need parentheses, but I'm not sure.  We also don't always need parentheses for the case `&dyn Trait`, but we do *sometimes* need it (e.g., in `dyn Fn() -> &dyn Trait + 'static` which is ambiguous), so I think it makes sense to always parenthesize a reference to a `dyn`/`impl Trait`.  This will make us have to keep track of much less, but it will not always match 1:1 with the source code because of the extraneous parentheses, but it will be semantically equivalent.

test_crates/raw_type_json/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+    c: Constant<25>,
+    d: impl for<'x> FnMut(
+            &'a unsafe extern "C" fn(
+                *const [u8],
+                &'x mut *mut (),
+                ...
+            ) -> std::borrow::Cow<'static, [u8]>,
+        ) -> &'x (dyn std::fmt::Display)
+        + Send
+        + 'static,
```
obi1kenobi, line 76: My eyes, they bleed! 😭

Jokes aside, I continue to be impressed by your attention to detail and to proper testing 💯

src/adapter/rust_type_name.rs
```diff
@@ -0,0 +1,676 @@
...
+            //
+            // Parentheses for an `impl/dyn Trait` with no `+` are not always required,
+            // but it can create ambiguity with e.g., `impl Fn() -> &dyn Trait + Send + Sync`,
+            // which could be `Fn() -> &(dyn Trait) + Send + Sync`
+            // or `Fn() -> &(dyn Trait + Send + Sync)`.
+            //
+            // Unconditionally wrapping an `impl/dyn Trait` reference with parentheses
+            // may create something syntactially different from the original statement,
+            // but it will always be semantically equivalent, and lets the formatting be more
+            // context-free than keeping track of when parentheses are needed.
```
obi1kenobi: Would this produce `&(dyn Trait)` notation though?

The eventual use of this code would be in materializing witness code, which we expect users will scrutinize and nitpick on. I'm not sure how much work it would be to only emit parens when strictly necessary, but it might be really good to look into it (possibly in a future PR) so that our generated Rust code looks as idiomatic as possible.

People will mistrust witnesses that they can't read and understand, so this is one place where paying a bit more development cost one time up front will probably pay off in the long run: many if not most of our witness generators will rely on this code.

src/adapter/rust_type_name.rs
```diff
@@ -0,0 +1,676 @@
...
+            Ok(())
+        }
+        rustdoc_types::Type::Tuple(tys) => {
+            write!(f, "(")?;
+            intersperse_with(f, ", ", tys, Type)?;
+            write!(f, ")")
+        }
+        rustdoc_types::Type::Slice(ty) => write!(f, "[{}]", Type(ty)),
+        // According to the docs for this variant, `len` is not guaranteed to be the
+        // same as the source code.  For example, [u8; 1 + 2] currently becomes [u8; 3]
```
obi1kenobi, line 243:
```suggestion
        // According to the docs for this variant, `len` is not guaranteed to be the
        // same as the source code.  For example, `[u8; 1 + 2]` currently becomes `[u8; 3]`.
```
Does this appear to be guaranteed to have the actual constant value present? E.g. (and I'm not sure if this is legal Rust) if we have `[u8; some_const_fn() + 1]` or something like that.

Or can it ever be `_` like the values of `const` items sometimes end up being in cases like `const ITEM: u8 = some_const_fn();` ?

src/adapter/rust_type_name.rs
obi1kenobi: I think that nothing attempts to write out the generic parameters of the function that holds `impl Trait`. The test case prints out the types of the function parameters (and finds `impl Trait`), but the synthetic generic param shows up in the function's own type itself instead of the thing being printed and tested.

Try adding an additional function parameter with type like `impl FnOnce(&impl Clone) -> impl Clone`. When trying to print that parameter's type, that will run into (I believe) two synthetic type parameters, one for `&impl Clone` and one for `-> impl Clone`.

Desugared, that type is more-or-less `<T: Clone> impl FnOnce(&T) -> Self::R where Self::R: Clone`.

obi1kenobi (commented):
Curious about your take on the parens issue 👇

If you think it's a ton of work and not worth it, we can skip it.

If you think it's doable, we can merge this as-is first and then improve it in the next PR.

We should probably merge this PR after the synthetic param issue is sorted out, so we don't have to keep re-reviewing on an 800 line PR over changes to 10 lines :)

obi1kenobi:
Retargeted this branch to target merging into `rustdoc-v36` instead of `rustdoc-v34`, and fixed an associated merge conflict.

src/adapter/rust_type_name.rs
suaviloquence: The docstring just says 

> Keep in mind that it's not guaranteed to match the actual source code of the expression.

```rust
const fn const_fn() -> usize {
    let mut i = 0;

    while i < 27 {
        i += 5;
    }

    i
}

// ...
    pub array: [(); 1 + 2],
    pub array2: [(); usize::max >> (usize::bits - 1)],
    pub array3: [(); const_fn()],
```

is serialized as:

```rs
array: [(); 3],
array2: [(); 1],
array3: [(); 30]
```

I don't know if there are any guarantees, but I added these test cases.

src/adapter/rust_type_name.rs
suaviloquence:
```
error[E0562]: `impl Trait` is not allowed in the parameters of `Fn` trait bounds
  --> src/lib.rs:53:17
   |
53 |     f: impl Fn(&impl Clone) -> impl Clone,
   |                 ^^^^^^^^^^
   |
   = note: `impl Trait` is only allowed in arguments and return types of functions and methods

error[E0562]: `impl Trait` is not allowed in the return type of `Fn` trait bounds
  --> src/lib.rs:53:32
   |
53 |     f: impl Fn(&impl Clone) -> impl Clone,
   |                                ^^^^^^^^^^
   |
   = note: `impl Trait` is only allowed in arguments and return types of functions and methods
```

It looks like that's not allowed.  I'm going to search the JSON for when `is_synthetic` is true.

suaviloquence: I figured it out.  This is only for `GenericParamDef`, which we only use when we have a `for<...>` bound.  This currently only supports lifetimes, so this is not expressible (a `for<T> Fn(T)` bound is not currently valid).  This may come up when we are serializing generic definitions in functions, though.  We could mark the whole type branch as `unreachable!` for now.

obi1kenobi: Yes, it's only in `GenericParamDef` which is why I was hoping to trigger it with `impl Fn(impl Trait)`, but I guess that won't work =/

I recently added [schema for generic parameters](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/514), among which some [test cases that do have synthetic generics](
https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/941f2e6a7cb6123f97da851ce710a4c1da17b41f/test_crates/generic_parameters/src/lib.rs#L24-L27).

A related problem we have right now is that we can't write the Rust representation of `ImplementedTrait` or `GenericTypeParameter`, which show up in trait bounds and in `impl<'a, T: Clone, const N: i64> SomeTrait<'a, T, N> for MyType`.

Marking this unreachable or todo for now is fine then. If you're open to tackling the `ImplementedTrait` / `GenericTypeParameter` type printing problem next, that would be lovely and super useful because it'll improve the cargo-semver-checks lint diagnostics right away: instead of saying "type no longer implements `SomeTrait`", we can say it "no longer implements `for<'a> SomeTrait<'a, T, 5> where T: Clone + 'a`" — because it might still implement some other, incompatible variation of `SomeTrait`!

src/adapter/rust_type_name.rs
obi1kenobi: Very nice. Having the test cases is great, so we'll know if this behavior changes in the future. Thanks!

obi1kenobi (approved):
How would you feel about merging this PR?

I think there are two threads that may be worth pulling on further: the parens, and the `ImplementedTrait` and `GenericTypeParameter` printing which will hit the `is_synthetic` branch. But both of those can go in their own PRs — I don't see a need to push them in here.

src/adapter/rust_type_name.rs
obi1kenobi: We'll have the equivalent problem with generic parameters: a future lint we'll have is "this impl used to only require `T: Clone` but now it's also `T: Sync`"
```rust
// before
impl<T: Clone> SomeType<T> {
    pub fn example(value: T) {}
}

// after
impl<T: Clone + Sync> SomeType<T> {
    pub fn example(value: T) {}
}
```
You can't call `SomeType::example` with `T: !Sync` anymore (e.g. with `Rc<i64>`), so this is breaking. We'll need to print the affected generic parameter and its bounds (lifetimes, `use`, and `ImplementedTrait`)

suaviloquence:
I would like to fix the parens issue, because it shows up in more cases than I thought, and I have most of a fix written. I think the `is_synthetic` branch will never occur unless (1) `for<T>` type HRTBs are added and stabilized (and whether you can even *use* synthetic `impl Trait` generics in this way) or (2) we use this code to print function generic definitions as well. The second case being more likely, I think I could address the is_synthetic issue in the PR that adds generic-definition-printing support.

suaviloquence:
I do think the commit diff view [here](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/528/commits/bde40e2f86af64bdd5b14225a816052eee3e4eda) will be helpful in reviewing these new changes.

Also, note that I'm using `dyn Trait` to mean `dyn Trait` or `impl Trait` here.

I added essentially a lookahead parameter to many of the wrapper types, for whether or not it should propagate "wrap `dyn Trait` in parentheses" to the rightmost type.  The rules seem to be:

1. If there is a reference `&dyn Trait` or raw pointer `*const dyn Trait`, it must be wrapped in parentheses (otherwise it will be rejected syntactically by the parser), regardless of other rules, if it contains multiple bounds with a `+` (this includes lifetime parameters in `dyn Trait`).  I do this with the `force_wrap_parens` variable in the `rustdoc_types::Type::{RawPointer,BorrowedRef}` branches in `fmt_type`.
2. if a `dyn Trait` is delimited by something other than `+`, it does not need to be wrapped in parentheses.  For example, in angle-brackets `Box<dyn Trait + ...>`, in a slice (probably more for `impl Trait`s that are `Sized), or in a tuple.
3. If a `dyn` or `impl` trait (or, typically a reference/raw pointer to one) is the *output type* of a function pointer or `impl/dyn Fn`, it *may* need to be wrapped in parentheses: if a `+` bound comes after the output type, this is ambiguous.  For example, `impl Fn() -> *const dyn Display + Send` is ambiguous and needs parentheses (where does the `Send` bound apply), but `impl Send + Fn() -> *const dyn Display` does not need parentheses.

Let me know if I can answer any more questions about this, I know this is probably not the easiest PR to review.

obi1kenobi (approved):

obi1kenobi:
Nice work! The code is quite decent overall, especially considering how complex the rules and the domain are. And I'm glad it's super well-tested, which gives me hope that it is correct as well :)

As one meta note: if you find yourself feeling that you should help the review process by writing an explanatory comment in the PR review conversation, consider pulling as much of that comment into the PR's source code as possible instead.

If something is confusing during review today, it'll probably be confusing when reading the code next month, when we'll be reading it _outside_ of the PR review screen. Even if we then remember to look up the discussion on the original PR, some of that discussion may be stale and the code may have been changed across multiple PRs. All this makes understanding the code much more difficult down the line, and makes everything less maintainable as a result.

Comments that are inline in the code, meanwhile, are less likely to drift. We review them as part of every new PR, and update them as needed. We don't update discussions in old, already-merged PRs. So prefer comments in the code over comments in the PR review discussion :)

PR #597, opened 2024-11-17
Add `signature` property on `FunctionLike`

suaviloquence:
This is my first time adding a property, so let me know if there is more to do/document/test.

Adds the `signature` property on `FunctionLike` vertices, and the `function_signature` method (and display wrappers as needed in `rust_type_name.rs`) to implement this.

### DIFF
--- a/src/adapter/tests.rs
+++ b/src/adapter/tests.rs
@@ -3635,7 +3635,7 @@ fn method_signature() {
             ... on Trait {
                 name @filter(op: "=", value: ["$trait"])
                 method {
-                    name @filter(op: "=", value: ["$method"]) @output
+                    name @output
                     signature @output
                 }
             }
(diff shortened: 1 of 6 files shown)


src/adapter/tests.rs
```diff
@@ -3564,3 +3564,111 @@ fn type_generic_bounds() {
...
+
+    let mut expected_results = vec![Output {
+        name: "method".into(),
+        signature: "fn method<'a, T, U: GAT<(T, ())>>() where Self: Sized,\n\
+                for<'b> <U as GAT<(T, ())>>::Type<'b, ()>: 'static"
+            .into(),
+    }];
+    expected_results.sort_unstable();
+    similar_asserts::assert_eq!(expected_results, results);
+}
```
obi1kenobi: Could you add a test case where both the method and its parent trait / struct etc. are generic, where the method's signature uses both sets of generic parameters?

I'm actually not sure about the best way to print that out. We can't append the parent's generics to the method's own generics, for example.

It might hint toward the need for a similar `signature` field on impls, structs, enums, unions, and traits that shows the generic parameters and bounds applied to that entire block.

src/rustdoc_schema.graphql
```diff
@@ -1198,6 +1198,11 @@ interface FunctionLike {
   unsafe: Boolean!
   async: Boolean!
   has_body: Boolean!
+  """
+  Stringified representation of the function's signature, as close to the
+  original source code as possible.
+  """
+  signature: String!
```
obi1kenobi: The description in the comment here seems to hint at deviations from the source code. Could you elaborate a bit on that? For example, it would be good to know if some property is never preserved, or if another one is always preserved, or maybe preserved and maybe not.

src/adapter/tests.rs
obi1kenobi: Another good test case: using a trait's associated type in the signature of one of that trait's own methods. It'd be interesting to pin down whether we print that as `Self::Type` or `<Self as Trait>::Type` for example.

Perhaps a similar `signature` field on associated types would be good too, to show their bounds.

obi1kenobi (approved):
This looks great, you nailed it.

I've just suggested another test case or two that would be good to pin down, and a small docs tweak in the schema that would make it easier to use. Happy to merge immediately and take those changes in a separate PR, if that's any easier on your end.

src/rustdoc_schema.graphql
suaviloquence: It's the same limitations as the typename, I think - constants won't always be printed identically (esp. with constant expressions like `A<{1 + 2}>`), we can't recreate whitespace (but we try to pretty-print it), and I'm not sure if paths are guaranteed to be the same (but they will point to the same item)

suaviloquence:
Added these fixes.  The `Self` type was a good catch - currently `rustdoc_types` encodes `Self::T` in a trait as `<Self as (empty string)>::T`, which is not desirable, so I fixed that

obi1kenobi (approved):

PR #598, opened 2024-11-18
Allow generating just specified rustdocs in `regenerate_test_rustdocs.sh`

suaviloquence:
Updates `regenerate_test_rustdocs.sh` to have the following behavior

1. if no args are passed (except an optional `+toolchain`), renerates all rustdocs in the `test_crates` directory
2. if the script is called with `regenerate_test_rustdocs.sh crate1 crate2 ...`, only regenerate `test_crates/crate1`, `test_crates/crate2`, etc.

### DIFF
--- a/scripts/regenerate_test_rustdocs.sh
+++ b/scripts/regenerate_test_rustdocs.sh
@@ -10,13 +10,27 @@ TARGET_DIR="$TOPLEVEL/localdata/test_data"
 
 # Allow setting an explicit toolchain, like +nightly or +beta.
 set +u
-TOOLCHAIN="$1"
+case "$1" in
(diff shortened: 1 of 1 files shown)


obi1kenobi (approved):
Nice!

If you want to make this even better, consider using the trick from the cargo-semver-checks test suite: check if the test crate's directory is newer than the rustdoc file (if it exists) — and if so then regenerate it because it's almost certainly stale:
https://github.com/obi1kenobi/cargo-semver-checks/blob/68611c445cbe98a2ae60cb0a0f5461f2ab201fab/scripts/regenerate_test_rustdocs.sh#L12-L16

PR #617, opened 2024-11-27
Do not return methods when querying top-level functions. (#614)

### LINKED ISSUE #400: Querying functions also produces methods, since at a glance they look identical in JSON
Spotted in #389, whose test case queries only free functions but also gets methods (associated functions) back. Its query is:
```graphql
{
    Crate {
        item {
            ... on Function {
                name @output
                has_body @output
            }
        }
    }
}
```
and incorrectly matches on associated functions like
```rust
pub struct Foo;

impl Foo {
    pub fn inside_impl_block() {}
}

pub trait Bar {
    fn trait_no_body();

    fn trait_with_body() {}
}
```

Many format versions ago, rustdoc JSON used to distinguish between `Function` and `Method` items. Associated functions were of type `Method` in the rustdoc JSON, so our code handling was correct. When the `Method` type got replaced with `Function`, we stopped being able to distinguish free and associated functions when iterating through the crate item index.

I [asked the rustdoc team](https://rust-lang.zulipchat.com/#narrow/stream/266220-t-rustdoc/topic/Denoting.20function.20items.20as.20associated.20or.20free.20in.20JSON/near/463718899) if they might be open to adding some way to distinguish free vs associated functions to rustdoc JSON. Otherwise, we'll need to add an index of our own that helps us avoid returning associated functions in a query for free functions like the above.

This isn't an immediate correctness concern for `cargo-semver-checks`, since associated functions have no instances of the `importable_path` edge and would get filtered out. It's probably just a performance concern, since we don't discard items from query processing as early as we could.

obi1kenobi:
Fixes #400 by adding an index of function item ID -> owning item i.e.
the struct, enum, union, or trait with which the function is associated.

### DIFF
--- a/src/adapter/optimizations/item_lookup.rs
+++ b/src/adapter/optimizations/item_lookup.rs
@@ -112,7 +112,20 @@ fn resolve_items_slow_path<'a>(
         .inner
         .index
         .values()
-        .filter(move |item| item.crate_id == own_crate_id);
+        .filter(move |item| item.crate_id == own_crate_id)
+        .filter(move |item| {
+            // We don't consider methods as top-level items in a crate.
+            // We should only return methods as owned items within a trait or impl.
+            // If we are to return this item, then either the item isn't a function at all,
+            // or it's a top-level function (i.e. has no owner listed in the index).
+            !matches!(item.inner, rustdoc_types::ItemEnum::Function(..))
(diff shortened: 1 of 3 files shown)


obi1kenobi:
Needs #621 to merge first, will retry in another couple of days.

PR #655, opened 2024-12-10
Bump the major versions of cargo_metadata and cargo_toml. (#650)

obi1kenobi:


Cargo.toml
```diff
@@ -13,14 +13,38 @@ readme = "./README.md"
...
+
+#
+# Keep this dependency last, its own block, since we change it often via automated means
+# and we'd like to avoid merge conflicts.
+#
+rustdoc-types = "0.33.0"
+#
+# No dependencies after this point. Put all dependencies above this section.
+#
+>>>>>>> baaf43e (Bump the major versions of cargo_metadata and cargo_toml. (#650))
```
obi1kenobi:
```suggestion
cargo_metadata = "0.19.1"
#
# Keep this dependency last, its own block, since we change it often via automated means
# and we'd like to avoid merge conflicts.
#
rustdoc-types = "0.28.0"
#
# No dependencies after this point. Put all dependencies above this section.
#
```

PR #676, opened 2024-12-17
Add a per-item-kind index of public items in the crate.

obi1kenobi:
Most items we'll see in the crate either aren't public, or aren't top-level items that would be returned by the `Crate -[item]-> Item` edge. For example, many of those items are methods, impls, trait associated types, associated constants, etc.

In this PR, we ensure we don't bother iterating over such items in queries where they can't possibly match. Queries that look at public structs should only iterate over structs, ones that look at enums should only iterate over enums, etc.

This has a substantial performance effect: it speeds up the median `cargo-semver-checks` lint by ~100x, and makes it so that a tiny handful (~5) of lints dominate all linting runtime — the combined runtime of all other lints becomes negligible. In end-to-end terms, `cargo-semver-checks` becomes about ~35% faster on our largest workload, but this number is misleadingly small w.r.t. the impact of this optimization: the new runtime is now equal to ~101% of the slowest lint, whereas that number was ~140% previously. If we can now optimize that slow lint in particular, we'll get tens of percent of additional speedup from the optimization in this PR.

### DIFF
--- a/src/adapter/optimizations/item_lookup.rs
+++ b/src/adapter/optimizations/item_lookup.rs
@@ -65,6 +65,195 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
         }
     }
 
+    // Is the `importable_path` edge being resolved in a subsequent step in a *mandatory* fashion?
+    // If so, we could only match on public items, so check which kinds of public items
+    // we're looking for and only return those from the index.
+    if destination
+        .first_mandatory_edge("importable_path")
+        .is_some()
+    {
+        if let Some(item_type) = destination.coerced_to_type().map(|x| x.as_ref()) {
+            match item_type {
+                "Function" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .free_functions
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "Struct" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .structs
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "Enum" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .enums
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "Union" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .unions
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "Trait" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .traits
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "ImplOwner" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .structs
+                                .iter()
+                                .chain(crate_vertex.pub_item_kind_index.enums.iter())
+                                .chain(crate_vertex.pub_item_kind_index.unions.iter())
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "Constant" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .free_consts
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "Static" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .statics
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "GlobalValue" => {
+                    // const or static
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .free_consts
+                                .iter()
+                                .chain(crate_vertex.pub_item_kind_index.statics.iter())
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    });
+                }
+                "Macro" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .decl_macros
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "ProcMacro"
+                | "FunctionLikeProcMacro"
+                | "AttributeProcMacro"
+                | "DeriveProcMacro" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .proc_macros
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                "Module" => {
+                    return resolve_neighbors_with(contexts, move |vertex| {
+                        let crate_vertex =
+                            vertex.as_indexed_crate().expect("vertex was not a Crate");
+                        let origin = vertex.origin;
+                        Box::new(
+                            crate_vertex
+                                .pub_item_kind_index
+                                .modules
+                                .iter()
+                                .map(move |item| origin.make_item_vertex(item)),
+                        )
+                    })
+                }
+                _ => {}
+            }
+        }
+    }
+
     resolve_neighbors_with(contexts, |vertex| {
         let crate_vertex = vertex.as_indexed_crate().expect("vertex was not a Crate");
         let origin = vertex.origin;
--- a/src/indexed_crate.rs
+++ b/src/indexed_crate.rs
@@ -198,6 +198,9 @@ pub struct IndexedCrate<'a> {
     /// -> function item with that export name; exported symbol names must be unique.
     pub(crate) export_name_index: Option<HashMap<&'a str, &'a Item>>,
 
+    /// index: kind of public top-level item -> `Vec<&'a Item>` of that kind
+    pub(crate) pub_item_kind_index: PubItemKindIndex<'a>,
+
     /// Trait items defined in external crates are not present in the `inner: &Crate` field,
     /// even if they are implemented by a type in that crate. This also includes
     /// Rust's built-in traits like `Debug, Send, Eq` etc.
@@ -212,6 +215,159 @@ pub struct IndexedCrate<'a> {
     /// A more complete future solution may generate multiple crates' rustdoc JSON
     /// and link to the external crate's trait items as necessary.
     pub(crate) manually_inlined_builtin_traits: HashMap<Id, Item>,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct PubItemKindIndex<'a> {
+    pub(crate) free_functions: Vec<&'a Item>,
+    pub(crate) structs: Vec<&'a Item>,
+    pub(crate) enums: Vec<&'a Item>,
+    pub(crate) unions: Vec<&'a Item>,
+    pub(crate) traits: Vec<&'a Item>,
+    pub(crate) modules: Vec<&'a Item>,
+    pub(crate) statics: Vec<&'a Item>,
+    pub(crate) free_consts: Vec<&'a Item>,
+    pub(crate) decl_macros: Vec<&'a Item>,
+    pub(crate) proc_macros: Vec<&'a Item>,
+}
+
+impl<'a> PubItemKindIndex<'a> {
+    fn with_capacity_hint(hint: usize) -> Self {
+        let capacity = if hint < 128 * 128 { 128 } else { hint / 128 };
+        Self {
+            // Most top-level items in a crate are functions, structs, enums, or traits.
+            // We want indexing to be fast, and it's okay if we waste a bit of memory.
+            free_functions: Vec::with_capacity(capacity),
+            structs: Vec::with_capacity(capacity),
+            enums: Vec::with_capacity(capacity),
+            traits: Vec::with_capacity(capacity),
+            unions: Vec::new(),
+            modules: Vec::with_capacity(64),
+            statics: Vec::new(),
+            free_consts: Vec::new(),
+            decl_macros: Vec::new(),
+            proc_macros: Vec::new(),
+        }
+    }
+
+    fn from_crate(crate_: &'a Crate, fn_owner_index: &HashMap<Id, &'a Item>) -> Self {
+        #[cfg(feature = "rayon")]
+        let iter = crate_.index.par_iter().map(|(_, value)| value);
+        #[cfg(not(feature = "rayon"))]
+        let iter = crate_.index.values();
+
+        #[cfg(feature = "rayon")]
+        let init = || Self::with_capacity_hint(crate_.index.len());
+        #[cfg(not(feature = "rayon"))]
+        let init = Self::with_capacity_hint(crate_.index.len());
+
+        let folded = iter.fold(init, |mut acc, item| {
+            if item.visibility == rustdoc_types::Visibility::Public {
+                match &item.inner {
+                    rustdoc_types::ItemEnum::Module { .. } => {
+                        acc.modules.push(item);
+                    }
+                    rustdoc_types::ItemEnum::Union { .. } => {
+                        acc.unions.push(item);
+                    }
+                    rustdoc_types::ItemEnum::Struct { .. } => {
+                        acc.structs.push(item);
+                    }
+                    rustdoc_types::ItemEnum::Enum { .. } => {
+                        acc.enums.push(item);
+                    }
+                    rustdoc_types::ItemEnum::Function { .. } => {
+                        if !fn_owner_index.contains_key(&item.id) {
+                            // This is a free function.
+                            acc.free_functions.push(item);
+                        }
+                    }
+                    rustdoc_types::ItemEnum::Trait { .. } => {
+                        acc.traits.push(item);
+                    }
+                    rustdoc_types::ItemEnum::Constant { .. } => {
+                        acc.free_consts.push(item);
+                    }
+                    rustdoc_types::ItemEnum::Static { .. } => {
+                        acc.statics.push(item);
+                    }
+                    rustdoc_types::ItemEnum::Macro { .. } => {
+                        acc.decl_macros.push(item);
+                    }
+                    rustdoc_types::ItemEnum::ProcMacro { .. } => {
+                        acc.proc_macros.push(item);
+                    }
+                    _ => {}
+                }
+            }
+
+            acc
+        });
+
+        #[cfg(feature = "rayon")]
+        let folded = {
+            let collected = folded
+                .map(|index| vec![index])
+                .reduce(Vec::new, |mut left, right| {
+                    left.extend(right);
+                    left
+                });
+
+            let mut fns_len = 0;
+            let mut structs_len = 0;
+            let mut enums_len = 0;
+            let mut unions_len = 0;
+            let mut traits_len = 0;
+            let mut modules_len = 0;
+            let mut statics_len = 0;
+            let mut consts_len = 0;
+            let mut decl_macros_len = 0;
+            let mut proc_macros_len = 0;
+
+            for c in &collected {
+                fns_len += c.free_functions.len();
+                structs_len += c.structs.len();
+                enums_len += c.enums.len();
+                unions_len += c.unions.len();
+                traits_len += c.traits.len();
+                modules_len += c.modules.len();
+                statics_len += c.statics.len();
+                consts_len += c.free_consts.len();
+                decl_macros_len += c.decl_macros.len();
+                proc_macros_len += c.proc_macros.len();
+            }
+
+            let mut value = Self {
+                free_functions: Vec::with_capacity(fns_len),
+                structs: Vec::with_capacity(structs_len),
+                enums: Vec::with_capacity(enums_len),
+                unions: Vec::with_capacity(unions_len),
+                traits: Vec::with_capacity(traits_len),
+                modules: Vec::with_capacity(modules_len),
+                statics: Vec::with_capacity(statics_len),
+                free_consts: Vec::with_capacity(consts_len),
+                decl_macros: Vec::with_capacity(decl_macros_len),
+                proc_macros: Vec::with_capacity(proc_macros_len),
+            };
+
+            for c in collected {
+                value.free_functions.extend(c.free_functions);
+                value.structs.extend(c.structs);
+                value.enums.extend(c.enums);
+                value.unions.extend(c.unions);
+                value.traits.extend(c.traits);
+                value.modules.extend(c.modules);
+                value.statics.extend(c.statics);
+                value.free_consts.extend(c.free_consts);
+                value.decl_macros.extend(c.decl_macros);
+                value.proc_macros.extend(c.proc_macros);
+            }
+
+            value
+        };
+
+        folded
+    }
 }
 
 /// Map a Key to a List (Vec) of values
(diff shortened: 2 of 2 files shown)


CLIDragon:
Four queries receive no speedup. Two of these (`feature_missing` and `feature_not_enabled_by_default`) deal with features and will not receive any speedup. The other two, `exported_function_changed_abi` and `function_export_name_changed` do not receive improvements because they deal with private items. 

The remaining slow queries and their runtimes from tracing on `aws-sdk-ec2` are in the table below. These 11 queries represent about three quarters of the total runtime. The numbers were generated from a single run, and are quite noisy; they are, however, illustrative of the relative runtimes between each query.
| Query                                          | Runtime (ms)   |
|------------------------------------------------|-----------|
| `auto_trait_impl_removed`                        | 1308 |
| `derive_trait_impl_removed`                      | 1544 |
| `enum_unit_variant_changed_kind`                 | 824  |
| `enum_variant_marked_non_exhaustive`             | 840  |
| `inherent_method_missing`                        | 1810 |
| `inherent_method_now_doc_hidden`                 | 1796 |
| `inherent_method_unsafe_added`                   | 2077 |
| `method_parameter_count_changed`                 | 2715 |
| `method_requires_different_const_generic_params` | 2425 |
| `method_requires_different_generic_type_params`  | 2432 |
| `partial_ord_enum_variants_reordered`            | 1146 |

obi1kenobi:
Just clarifying, the runtimes are _before_ the optimization is applied, right?

Also, are the numbers from a single run, or a derived number (e.g. median) across multiple runs? How stable are they between runs? I'm asking because you reported 4 decimal places, and I'm not confident we actually _have_ valid measurement data on the behavior of the 4th decimal place — I think it's actually just noise. But I'm not sure if it's noise all the way down to seconds or tenths or hundredths.

CLIDragon:
The runtimes are, in ms, from a single run after optimisation, but including the overhead from perf-tracing. I don't have exact numbers for inter-run noise, but the actual numbers don't really matter so much as the relationship between them. I've edited the above post to reflect this. For reference, the mean runtime of all other queries (excluding the 11 longest) is 35ms, and the median is 1.76ms - i.e. they are in the order of 20-75x greater than the mean, and over a 1000x longer than the median query.

obi1kenobi:
If you're open to it, I think it's worth a look at all the lints with runtimes above 2000ms in that table.

All of them are written very defensively, making assumptions about how the Rust compiler might attempt to "interpret the code to make it correct." I'm not necessarily convinced all those assumptions are true  — they may be overly conservative. If that is the case, we may be able to simplify the queries and get an additional perf improvement.

For example, the `inherent_method_unsafe_added` lint is written to assume that:
- if a safe inherent method by the given name used to exist
- if that method has since become unsafe, but a trait implemented on that type still has a safe method by the same name
- that `rustc` will then "automatically" choose to use the safe method from the trait since the `unsafe` inherent one would require an `unsafe` block.

I'm not sure that's true! To make an improvement here, we'd first need to tighten the test crate for this lint with a test for this specific case. We can use a doctest to prove the _actual_ behavior of `rustc` and guard against regressions in future Rust versions: recall that doctests are compiled as a separate crate, so we can use them to see how "a downstream crate" experiences the public API of the crate. I also believe we already run `cargo test` on all test crates, but that'd be good to check.

A similar pass over the other lints would be good to prove or disprove the need to be as conservative (and complex) in those queries.

I believe we can probably cut on the order of 0.5-1s off of each of those lints, which will obviously have a large positive impact on overall runtime since they are the long poles in the perf graph.

How does that sound as a next step?

obi1kenobi:
I'd also be curious to know, for those four lints, for the three consecutive `impl @fold` blocks in the defensively-written portion of the query:
- how many method items each one sees
- how much time is spent in each

This is the performance anomaly I had been referring to, which I just realized might not have been obvious thus far. I would have expected these lints to benefit more here, and I'm not entirely sure why they don't seem to have.

CLIDragon:
I dug into `method_parameter_count_changed` and there is definitely an issue. Below is the initial callstack (the calls that appear at the top of a trace) followed by the number of times the `Opid` appears in the trace square brackets. Before applying this PR, `Opid(4)` has 4711 entries, so the index clearly reduces the number of calls there, but there is no change in the occurrences for `Opid(16)`.

The problem is that a path is resolved later so the first branch is triggered instead of the item-kind index. However, the first branch is significantly faster because it resolves the path. The best solution is to combine the results of the two indices.

```
Opid(1) ResolveStartingVertices(Vid(1))
Opid(2) ResolveNeighbors(Vid(1), "CrateDiff", Eid(1))
Opid(3) ResolveNeighbors(Vid(2), "Crate", Eid(2))
Opid(4) ResolveCoercion(Vid(3), "Item", "ImplOwner")  [196]
Opid(5) ResolveProperty(Vid(3), "ImplOwner", "visibility_limit")  [196]
Opid(6) ResolveNeighbors(Vid(3), "ImplOwner", Eid(3))  [196]
Opid(7) ResolveProperty(Vid(4), "ImportablePath", "public_api")  [196]
Opid(8) ResolveNeighbors(Vid(3), "ImplOwner", Eid(4))  [196]
Opid(9) ResolveNeighbors(Vid(5), "Impl", Eid(5))  [124]
Opid(10) ResolveProperty(Vid(6), "Method", "visibility_limit")  [142]
Opid(11) ResolveProperty(Vid(6), "Method", "public_api_eligible")  [142]
Opid(12) ResolveNeighbors(Vid(6), "Method", Eid(6))  [142]
Opid(13) ResolveNeighbors(Vid(6), "Method", Eid(7))  [142]
Opid(14) ResolveNeighbors(Vid(1), "CrateDiff", Eid(8))  [142]
Opid(15) ResolveNeighbors(Vid(9), "Crate", Eid(9))  [142]
Opid(16) ResolveCoercion(Vid(10), "Item", "ImplOwner")  [2386]  <-- Optimisation does not apply
Opid(17) ResolveProperty(Vid(10), "ImplOwner", "visibility_limit")  [142]
Opid(18) ResolveNeighbors(Vid(10), "ImplOwner", Eid(10))  [142]
Opid(19) ResolveProperty(Vid(11), "ImportablePath", "path")  [142]
Opid(20) ResolveProperty(Vid(4), "ImportablePath", "path")  [142]
Opid(21) ResolveProperty(Vid(11), "ImportablePath", "public_api")  [142]
Opid(22) ResolveProperty(Vid(6), "Method", "name")  [142]
Opid(23) ResolveNeighbors(Vid(10), "ImplOwner", Eid(11))  [142]
Opid(24) ResolveProperty(Vid(6), "Method", "name")  [142]
Opid(25) ResolveNeighbors(Vid(10), "ImplOwner", Eid(13))  [142]
Opid(26) ResolveProperty(Vid(6), "Method", "name")
Opid(27) ResolveNeighbors(Vid(10), "ImplOwner", Eid(16))
Opid(28) ResolveProperty(Vid(6), "Method", "name")
Opid(29) ResolveProperty(Vid(6), "Method", "visibility_limit")
Opid(30) ResolveProperty(Vid(10), "ImplOwner", "name")
Opid(31) ResolveProperty(Vid(4), "ImportablePath", "path")
Opid(32) ResolveProperty(Vid(8), "Span", "begin_line")
Opid(33) ResolveProperty(Vid(8), "Span", "end_line")
Opid(34) ResolveProperty(Vid(8), "Span", "filename")
Opid(35) ResolveProperty(Vid(3), "ImplOwner", "visibility_limit")
```

CLIDragon:
Unfortunately, because both indexes return vectors it costs more time to construct an intermediate `HashSet` than is saved by removing redundant operations.

obi1kenobi:
These two lines are very interesting, and anomalous:
```
Opid(15) ResolveNeighbors(Vid(9), "Crate", Eid(9))  [142]
Opid(16) ResolveCoercion(Vid(10), "Item", "ImplOwner")  [2386]  <-- Optimisation does not apply
```

If there are 142 matching `ImplOwner (at some importable path) -> Method` pairs before that line, we would expect only 142 item lookups because of our "look up by importable path" optimization. We seem to get that in `Opid(15)`.

I would then expect to have 142 coercion operations at `Opid(16)`. Each coercion operates on a single vertex and makes a boolean decision: either it lets the vertex through if it's of the right type, or it discards it otherwise.

I'm confused why we seem to be performing more than 16x more coercions than there are input vertices produced by the previous operation, `Opid(15)`. More interestingly, we seem to be back to 142 operations right afterward at `Opid(17)`.

There's definitely something unexpected happening here, and it's worth figuring it out. But I wanted to double-check: is figuring out this mystery something you find motivating and fun, or does it feel tedious instead? I'm very happy to let you drive this forward if you're excited about it, and I'm also happy to let you jump to making additional optimizations elsewhere if you're bored of it and you'd prefer me to take over :) No pressure either way — just making sure the work isn't a grind.

CLIDragon:
I have already found the cause, but I would appreciate any guidance on how to efficiently resolve it. There are two branches inside `resolve_crate_items` before the `pub_item_kind_index` is accessed. Because the query matches the path, the first branch (line 51`item_lookup.rs`) triggers and so `Opid(15)` returns all items with the given path. This is then reduced down in `Opid(16)` to only the `ImplOwner`s. The solution is to somehow combine the two indices - `imports_index` and `pub_item_kind_index` so that only the 142 matching `ImplOwner`s with the correct path are returned from `Opid(15)`.

CLIDragon:
The overhead of constructing a `HashSet` is too much to do it inside the loop (e.g. in `resolve_items_by_importable_path_field_value`), so I am currently looking at storing a `IndexMap<(id, crate_id), &'a Item>` instead of a `Vec` inside `PubItemKindIndex`.

obi1kenobi:
> all items with the given path

Are there really 16x more items at the `ImplOwner`s paths though? This is what's really suspicious to me. Items can be at the same path in different namespaces, sure, but there aren't 16 namespaces to go around. We can have a type (struct/enum/union etc.) and e.g. a derive macro at the same path, but even finding a function at that same path would be strange because it would have to have a name in uppercase like a type. We're still short 13 :)

So I'm not sure we've _fully_ found out what's going on here.

CLIDragon:
> Are there really 16x more items at the `ImplOwner`s paths though? This is what's really suspicious to me. Items can be at the same path in different namespaces, sure, but there aren't 16 namespaces to go around. We can have a type (struct/enum/union etc.) and e.g. a derive macro at the same path, but even finding a function at that same path would be strange because it would have to have a name in uppercase like a type. We're still short 13 :)

The numbers reported are raw appearances inside the trace file. This is a somewhat unintuitive metric - the actual number of calls is the number reported is given by $\text{calls} = \frac{n - 4}{3}$. There are 46 valid items in that trace, and 794 calls of `Opid(16)`. 

Each of the 46 items has a different number of items for their path. For example, the path `"serde", "de", "impls", "BTreeMapVisitor"` has the following 16 items:
```
Path: ["serde", "de", "impls", "BTreeMapVisitor"]
(Item { id: Id(645), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "T", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "U", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("U"), bounds: [TraitBound { trait_: Path { path: "TryFrom", id: Id(32), args: Some(AngleBracketed { args: [Type(Generic("T"))], constraints: [] }) }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "TryInto", id: Id(36), args: Some(AngleBracketed { args: [Type(Generic("U"))], constraints: [] }) }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(31), Id(33)], is_negative: false, is_synthetic: false, blanket_impl: Some(Generic("T")) }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(641), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "T", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("T"), bounds: [TraitBound { trait_: Path { path: "Sized", id: Id(16), args: None }, generic_params: [], modifier: Maybe }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "Borrow", id: Id(17), args: Some(AngleBracketed { args: [Type(Generic("T"))], constraints: [] }) }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(14)], is_negative: false, is_synthetic: false, blanket_impl: Some(Generic("T")) }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(643), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "T", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "U", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("U"), bounds: [TraitBound { trait_: Path { path: "From", id: Id(25), args: Some(AngleBracketed { args: [Type(Generic("T"))], constraints: [] }) }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "Into", id: Id(28), args: Some(AngleBracketed { args: [Type(Generic("U"))], constraints: [] }) }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(26)], is_negative: false, is_synthetic: false, blanket_impl: Some(Generic("T")) }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(632), crate_id: 0, name: Some("BTreeMapVisitor"), span: Some(Span { filename: "src\\de\\impls.rs", begin: (714, 1), end: (721, 22) }), visibility: Public, docs: Some("A visitor that produces a map."), links: {}, attrs: [], deprecation: None, inner: Struct(Struct { kind: Plain { fields: [], has_stripped_fields: true }, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [] }, impls: [Id(634), Id(635), Id(636), Id(637), Id(638), Id(639), Id(640), Id(641), Id(642), Id(643), Id(644), Id(645), Id(646), Id(647), Id(653)] }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(637), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [] }, provided_trait_methods: [], trait_: Some(Path { path: "Freeze", id: Id(7), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [], is_negative: false, is_synthetic: true, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(638), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [] }, provided_trait_methods: [], trait_: Some(Path { path: "Unpin", id: Id(9), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [], is_negative: false, is_synthetic: true, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(647), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "T", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("T"), bounds: [Outlives("'static"), TraitBound { trait_: Path { path: "Sized", id: Id(16), args: None }, generic_params: [], modifier: Maybe }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "Any", id: Id(48), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(45)], is_negative: false, is_synthetic: false, blanket_impl: Some(Generic("T")) }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(653), crate_id: 0, name: None, span: Some(Span { filename: "src\\de\\impls.rs", begin: (714, 1), end: (721, 22) }), visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("K"), bounds: [TraitBound { trait_: Path { path: "Deserialize", id: Id(161), args: None }, generic_params: [], modifier: None }, TraitBound { trait_: Path { path: "Eq", id: Id(59), args: None }, generic_params: [], modifier: None }, TraitBound { trait_: Path { path: "Ord", id: Id(76), args: None }, generic_params: [], modifier: None }], generic_params: [] }, BoundPredicate { type_: Generic("V"), bounds: [TraitBound { trait_: Path { path: "Deserialize", id: Id(161), args: None }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: ["visit_bool", "visit_isize", "visit_i8", "visit_i16", "visit_i32", "visit_i64", "visit_usize", "visit_u8", "visit_u16", "visit_u32", "visit_u64", "visit_f32", "visit_f64", "visit_char", "visit_str", "visit_string", "visit_unit", "visit_unit_struct", "visit_none", "visit_some", "visit_newtype_struct", "visit_seq", "visit_map", "visit_bytes", "visit_byte_buf"], trait_: Some(Path { path: "Visitor", id: Id(188), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(648), Id(650), Id(651)], is_negative: false, is_synthetic: false, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(642), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "T", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("T"), bounds: [TraitBound { trait_: Path { path: "Sized", id: Id(16), args: None }, generic_params: [], modifier: Maybe }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "BorrowMut", id: Id(20), args: Some(AngleBracketed { args: [Type(Generic("T"))], constraints: [] }) }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(18)], is_negative: false, is_synthetic: false, blanket_impl: Some(Generic("T")) }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(639), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("K"), bounds: [TraitBound { trait_: Path { path: "RefUnwindSafe", id: Id(13), args: None }, generic_params: [], modifier: None }], generic_params: [] }, BoundPredicate { type_: Generic("V"), bounds: [TraitBound { trait_: Path { path: "RefUnwindSafe", id: Id(13), args: None }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "UnwindSafe", id: Id(11), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [], is_negative: false, is_synthetic: true, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(636), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("K"), bounds: [TraitBound { trait_: Path { path: "Sync", id: Id(5), args: None }, generic_params: [], modifier: None }], generic_params: [] }, BoundPredicate { type_: Generic("V"), bounds: [TraitBound { trait_: Path { path: "Sync", id: Id(5), args: None }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "Sync", id: Id(5), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [], is_negative: false, is_synthetic: true, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(646), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "T", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "U", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("U"), bounds: [TraitBound { trait_: Path { path: "Into", id: Id(28), args: Some(AngleBracketed { args: [Type(Generic("T"))], constraints: [] }) }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "TryFrom", id: Id(32), args: Some(AngleBracketed { args: [Type(Generic("U"))], constraints: [] }) }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(37), Id(39)], is_negative: false, is_synthetic: false, blanket_impl: Some(Generic("T")) }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(634), crate_id: 0, name: None, span: Some(Span { filename: "src\\de\\impls.rs", begin: (714, 1), end: (721, 22) }), visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [] }, provided_trait_methods: [], trait_: None, for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(633)], is_negative: false, is_synthetic: false, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(635), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("K"), bounds: [TraitBound { trait_: Path { path: "Send", id: Id(3), args: None }, generic_params: [], modifier: None }], generic_params: [] }, BoundPredicate { type_: Generic("V"), bounds: [TraitBound { trait_: Path { path: "Send", id: Id(3), args: None }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "Send", id: Id(3), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [], is_negative: false, is_synthetic: true, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(640), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "K", kind: Type { bounds: [], default: None, is_synthetic: false } }, GenericParamDef { name: "V", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [BoundPredicate { type_: Generic("K"), bounds: [TraitBound { trait_: Path { path: "RefUnwindSafe", id: Id(13), args: None }, generic_params: [], modifier: None }], generic_params: [] }, BoundPredicate { type_: Generic("V"), bounds: [TraitBound { trait_: Path { path: "RefUnwindSafe", id: Id(13), args: None }, generic_params: [], modifier: None }], generic_params: [] }] }, provided_trait_methods: [], trait_: Some(Path { path: "RefUnwindSafe", id: Id(13), args: None }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [], is_negative: false, is_synthetic: true, blanket_impl: None }) }, Modifiers { doc_hidden: false, deprecated: false }), 
(Item { id: Id(644), crate_id: 0, name: None, span: None, visibility: Default, docs: None, links: {}, attrs: [], deprecation: None, inner: Impl(Impl { is_unsafe: false, generics: Generics { params: [GenericParamDef { name: "T", kind: Type { bounds: [], default: None, is_synthetic: false } }], where_predicates: [] }, provided_trait_methods: [], trait_: Some(Path { path: "From", id: Id(25), args: Some(AngleBracketed { args: [Type(Generic("T"))], constraints: [] }) }), for_: ResolvedPath(Path { path: "BTreeMapVisitor", id: Id(632), args: Some(AngleBracketed { args: [Type(Generic("K")), Type(Generic("V"))], constraints: [] }) }), items: [Id(29)], is_negative: false, is_synthetic: false, blanket_impl: Some(Generic("T")) }) }, Modifiers { doc_hidden: false, deprecated: false })])
```

It looks like the index is keeping every trait implementation for the path, which explains where the 16x factor is coming from.

CLIDragon:
I think I have a working solution. In particular, here is the runtimes of some of the longest queries, before and after. These are from a single run of c-s-c in verbose mode, so there is an uncertainty of at least 200ms in each value. Regardless, there is a clear trend in reducing runtime and operations. 
| Before   |Before|After| After      |                                                |
|----------|------------|----------|------------|------------------------------------------------|
| Time (s) | Operations | Time (s) | Operations | Query                                          |
| 1.062    | 10015524   | 0.913    | 6667432    | auto_trait_impl_removed                        |
| 1.63     | 12082820   | 1.193    | 10017052   | derive_trait_impl_removed                      |
| 0.997    | 2275338    | 0.797    | 1800770    | enum_variant_missing                           |
| 1.238    | 13016116   | 0.863    | 3864880    | inherent_method_missing                        |
| 1.08     | 12872087   | 0.635    | 3720851    | inherent_method_now_doc_hidden                 |
| 1.259    | 14688770   | 0.954    | 5537534    | inherent_method_unsafe_added                   |
| 2.255    | 14950044   | 1.264    | 5798808    | method_parameter_count_changed                 |
| 1.689    | 15285772   | 1.11     | 6134536    | method_requires_different_const_generic_params |
| 1.687    | 15761063   | 1.103    | 6609827    | method_requires_different_generic_type_params  |
| 1.103    | 7560367    | 1.195    | 7164863    | partial_ord_enum_variants_reordered            |
| 0.829    | 2837336    | 0.756    | 2099404    | sized_impl_removed                             |

obi1kenobi:
Merged as part of #902. Only took about 8 months 😅

Issue #535, opened 2024-10-18
Change trait "object safe" terminology to "dyn compatible"


obi1kenobi:
Rust more broadly is adopting this terminology change: https://github.com/rust-lang/rust/issues/130852

We should update our `Trait` data model to use the new terminology in its properties.

To avoid hard breakage, this should be a two-step process:
- [x] add a new `dyn_compatible: Boolean!` property that has identical data to the existing `object_safe: Boolean!` property
- [ ] eventually, after a suitable migration period, remove the `object_safe` property

PR #718, opened 2025-01-27
add `dyn_compatible` field to `Trait` data model

Frank-III:
close part of #535

### DIFF
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -446,10 +446,7 @@ pub(super) fn resolve_trait_property<'a, V: AsVertex<Vertex<'a>> + 'a>(
 ) -> ContextOutcomeIterator<'a, V, FieldValue> {
     match property_name {
         "unsafe" => resolve_property_with(contexts, field_property!(as_trait, is_unsafe)),
-        "object_safe" => {
-            resolve_property_with(contexts, field_property!(as_trait, is_dyn_compatible))
-        }
-        "dyn_compatible" => {
+        "object_safe" | "dyn_compatible" => {
             resolve_property_with(contexts, field_property!(as_trait, is_dyn_compatible))
         }
         "sealed" => resolve_property_with(contexts, move |vertex| {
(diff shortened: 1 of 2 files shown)


src/adapter/properties.rs
```diff
@@ -449,6 +449,9 @@ pub(super) fn resolve_trait_property<'a, V: AsVertex<Vertex<'a>> + 'a>(
         "object_safe" => {
             resolve_property_with(contexts, field_property!(as_trait, is_dyn_compatible))
         }
+        "dyn_compatible" => {
+            resolve_property_with(contexts, field_property!(as_trait, is_dyn_compatible))
+        }
```
obi1kenobi: We can merge these blocks to make sure they remain identical.
```suggestion
        "object_safe" | "dyn_compatible" => {
            resolve_property_with(contexts, field_property!(as_trait, is_dyn_compatible))
        }
```

obi1kenobi (approved):
Thank you!

I'll port this to the other active adapter versions (for different rustdoc JSON formats) using my automation scripts.

In the meantime, would you care to open a PR updating all cargo-semver-checks lints to use `dyn_compatible` instead of `object_safe`? Its tests will fail initially until this is published, but I'll take care of it from there and will merge it post-publish.

obi1kenobi:
Btw, not sure if you're already there so just in case: we have a Discord you're welcome to join if you'd like! Here's an invite link: https://discord.gg/WKWHFDCa

Frank-III:
> Thank you!
> 
> I'll port this to the other active adapter versions (for different rustdoc JSON formats) using my automation scripts.
> 
> In the meantime, would you care to open a PR updating all cargo-semver-checks lints to use `dyn_compatible` instead of `object_safe`? Its tests will fail initially until this is published, but I'll take care of it from there and will merge it post-publish.

I would give it a try!

obi1kenobi:
Another good first issue is #724, which then naturally enables another lint in `cargo-semver-checks` which we'd call something like `pub_safe_static_now_unsafe`. The gist of that lint is that a static that previously was accessible without an `unsafe {}` block now requires such a block, breaking existing code.

PR #737, opened 2025-01-28
Add export name to static

### LINKED ISSUE #238: Support `#[no_mangle]` and `#[export_name]` on statics
In #237 we added an `export_name` property on `Function`.

`Static` should get the same property, computed in the same way.

Frank-III:
close #238.
But in Rust version 2024: https://doc.rust-lang.org/reference/abi.html?highlight=no_mang#the-no_mangle-attribute, we need to use `#[unsafe(no_mangle)]` and `#[unsafe(export_name = "name")]`. Should we account for that?

### DIFF
--- a/src/adapter/properties.rs
+++ b/src/adapter/properties.rs
@@ -274,7 +274,7 @@ pub(super) fn resolve_function_property<'a, V: AsVertex<Vertex<'a>> + 'a>(
     match property_name {
         "export_name" => resolve_property_with(contexts, move |vertex| {
             let item = vertex.as_item().expect("not an Item vertex");
-            crate::exported_name::function_export_name(item).into()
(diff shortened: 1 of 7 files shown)


obi1kenobi:
Great catch — yes, if possible, add a new `edition = 2024` test crate that uses the new syntax. Let's make sure it gets parsed correctly in those cases as well, on both functions and statics.

Our current setup for building test crates' rustdoc enables nightly functionality, so I think `edition = 2024` should work (at worst) with a feature opt-in attribute at the top of the `lib.rs` file.

obi1kenobi:
This PR is great as-is, so I'll merge it. The edition-related changes can go in a new PR if that's alright, just to keep each PR focused on one thing at a time.

PR #764, opened 2025-02-07
feat: Add position property to StructField, Variant, and subtypes

### LINKED ISSUE #758: Add `position` property to `StructField`, `Variant`, and subtypes
The `position` is a 1-based numerical index showing the declaration order of the items in question. For example:
```rust
enum Example {
    First,
    Second,
}
```
Here `First` has `position = 1`, and `Second` has `position = 2`.

Position information is relevant because field and variant reordering can be a major breaking change. For example:
```rust
#[repr(C)]
pub struct A {
    x: i64,
    y: [f64; 2],
}

#[repr(C)]
pub struct B {
    y: [f64; 2],
    x: i64,
}
```
These two structs are _guaranteed_ to have different memory layouts, thanks to `repr(C)`. This means the two definitions are _not_ interchangeable in an FFI use case, and switching from one to the other would be a major breaking change!

Frank-III:
close: #758 

it seems we have to add another vertex kind to get the struct field position? let me know if there is a better way to do it!

I would add tests for struct fields, union fields, enum struct variants fields/tuple variants fields/plain variants/unit variants if we think the code is good


obi1kenobi (commented):
This approach is fine, but might get a bit repetitive and cause a bunch of refactoring churn.

Another approach might be to define a `VertexKind` variant like `VertexKind::PositionedItem(usize, &'a Item)` for all the things that are currently `VertexKind::Item` but also should get a position. That would include both struct fields and enum variants, and anything else that in the future might get a position index too. Then we could even make `VertexKind::as_item()` return the item for `PositionedItem` too, to minimize other code changes.

I'd also store the 1-indexed position in the item instead of doing a `+ 1` every time it's accessed, but that's minor and probably just personal taste.

Frank-III:
> Another approach might be to define a `VertexKind` variant like `VertexKind::PositionedItem(usize, &'a Item)` for all the things that are currently `VertexKind::Item` but also should get a position. That would include both struct fields and enum variants, and anything else that in the future might get a position index too. Then we could even make `VertexKind::as_item()` return the item for `PositionedItem` too, to minimize other code changes.

I tried to port all possible items into positioned items, but I left `EnumVariant` as it is and crate root_module as an item still, others seem all could be positioned items I suppose? Also, for items like `associated_constant` in `Impl` we want the position to be consecutive right, which means we need to do `filter_map` and then another `map`?

> I'd also store the 1-indexed position in the item instead of doing a `+ 1` every time it's accessed, but that's minor and probably just personal taste.

agreed!

obi1kenobi:
I wouldn't recommend porting all possible items into positioned items _proactively_. That constrains our future choices (e.g. making future refactoring more complex), at no additional benefit today.

I'd recommend two things:
- port the ones where we need that functionality now
- make it possible to, _should we need it in the future_, port other ones with minimum fuss.

Frank-III:
That make sense, another lesson learnt

obi1kenobi (commented):
This looks good! If you could add some tests for all the new functionality, I'd be happy to merge & publish a new release.

obi1kenobi (approved):

PR #806, opened 2025-03-10
Add schema and test cases for method receivers.

obi1kenobi:
Needs implementation in the adapter — tests are failing because the schema declares items that aren't implemented in the adapter yet.

PR #807, opened 2025-03-14
feat: add method `receiver` edge

Frank-III:
finish on top of: #806 

one thing I need to dive into would be that we wrap a lot of things in a `Rc`, is it because of how trustfall engine works, the examples there seem to do the same, but I think we don't need this for the `Receiver`?

### DIFF
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -12,6 +12,7 @@ use crate::{adapter::supported_item_kind, attributes::Attribute, PackageIndex};
 
 use super::{
     enum_variant::LazyDiscriminants,
+    method_self_receiver::MethodSelfReceiver,
     optimizations,
     origin::Origin,
     vertex::{Feature, Vertex},
@@ -258,6 +259,35 @@ pub(super) fn resolve_function_like_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
             Box::new(std::iter::once(origin.make_function_abi_vertex(abi)))
         }),
         _ => unreachable!("resolve_function_like_edge {edge_name}"),
+    }
+}
+
+pub(super) fn resolve_method_receiver_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
+    contexts: ContextIterator<'a, V>,
+    edge_name: &str,
+) -> ContextOutcomeIterator<'a, V, VertexIterator<'a, Vertex<'a>>> {
+    match edge_name {
+        "receiver" => resolve_neighbors_with(contexts, move |vertex| {
+            let origin = vertex.origin;
+            let method = vertex.as_function().expect("vertex was not a Function");
+
+            // Check if the first parameter is a self receiver
+            let receiver = method.sig.inputs.first().and_then(|(name, ty)| {
+                if name.starts_with("self") {
+                    Some(MethodSelfReceiver::new(ty))
+                } else {
+                    None
+                }
+            });
+
+            match receiver {
+                Some(receiver) => Box::new(std::iter::once(
+                    origin.make_method_receiver_vertex(receiver),
+                )),
+                None => Box::new(std::iter::empty()),
+            }
+        }),
+        _ => unreachable!("resolve_method_receiver_edge {edge_name}"),
     }
 }
 
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -262,7 +262,7 @@ pub(super) fn resolve_function_like_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
     }
 }
 
-pub(super) fn resolve_method_receiver_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
+pub(super) fn resolve_receiver_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
     contexts: ContextIterator<'a, V>,
     edge_name: &str,
 ) -> ContextOutcomeIterator<'a, V, VertexIterator<'a, Vertex<'a>>> {
@@ -286,7 +286,7 @@ pub(super) fn resolve_method_receiver_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
                     .map(move |r| origin.make_receiver_vertex(r)),
             )
         }),
-        _ => unreachable!("resolve_method_receiver_edge {edge_name}"),
+        _ => unreachable!("resolve_receiver_edge {edge_name}"),
     }
 }
 
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -12,9 +12,9 @@ use crate::{adapter::supported_item_kind, attributes::Attribute, PackageIndex};
 
 use super::{
     enum_variant::LazyDiscriminants,
-    method_self_receiver::MethodSelfReceiver,
     optimizations,
     origin::Origin,
+    receiver::Receiver,
     vertex::{Feature, Vertex},
     RustdocAdapter,
 };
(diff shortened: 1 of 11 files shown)


src/adapter/edges.rs
```diff
@@ -261,6 +262,35 @@ pub(super) fn resolve_function_like_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
...
+) -> ContextOutcomeIterator<'a, V, VertexIterator<'a, Vertex<'a>>> {
+    match edge_name {
+        "receiver" => resolve_neighbors_with(contexts, move |vertex| {
+            let origin = vertex.origin;
+            let method = vertex.as_function().expect("vertex was not a Function");
+
+            // Check if the first parameter is a self receiver
+            let receiver = method.sig.inputs.first().and_then(|(name, ty)| {
+                if name.starts_with("self") {
+                    Some(MethodSelfReceiver::new(ty))
```
obi1kenobi: Can you add a test case where a method's first argument is named e.g. `selfless` and see how that behaves? Because that shouldn't be a receiver, I think, and I suspect this will treat it as one.

src/adapter/edges.rs
```diff
@@ -261,6 +262,35 @@ pub(super) fn resolve_function_like_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
...
+                    None
+                }
+            });
+
+            match receiver {
+                Some(receiver) => Box::new(std::iter::once(
+                    origin.make_method_receiver_vertex(receiver),
+                )),
+                None => Box::new(std::iter::empty()),
+            }
```
obi1kenobi: There's a cute trick you can do here: `Option<T>` has [`.into_iter()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.into_iter) so you can do:
```suggestion
            Box::new(receiver.into_iter().map(|r| origin.make_method_receiver_vertex(r)))
```

src/adapter/method_self_receiver.rs
```diff
@@ -0,0 +1,96 @@
...
+        matches!(
+            self.0,
+            Type::BorrowedRef {
+                is_mutable: true,
+                ..
+            }
+        )
+    }
+
+    pub(super) fn kind(&self) -> String {
```
obi1kenobi: Let's have this and its helper function both return `Cow<'_, str>` instead, so that we don't have to always copy and reallocate strings when the exact string can be borrowed from `self.0`?

src/adapter/vertex.rs
```diff
@@ -59,6 +59,9 @@ pub enum VertexKind<'a> {
     #[non_exhaustive]
     FunctionAbi(&'a Abi),
 
+    #[non_exhaustive]
+    MethodReceiver(MethodSelfReceiver<'a>),
```
obi1kenobi: nit: The naming is pretty inconsistent from place to place.
- In the schema, the type is called `Receiver`.
- The variant is `MethodReceiver`.
- The internal type is `MethodSelfReceiver`.

I could possibly be convinced to have two different names instead of just `Receiver` everywhere. But three names for the same thing is definitely not what we want.

A priori, I still feel that `Receiver` is probably best as a one and only name, because while it's a bit rarely used in Rust, it's unambiguous. For example, a Google search for "receiver rust" returns info on the channel receiver type (obviously not a match) and [method receivers](https://google.github.io/comprehensive-rust/methods-and-traits/methods.html). But if you feel strongly about having different names in the schema and the code, let me know your thoughts — I could be convinced otherwise.

src/adapter/properties.rs
```diff
@@ -306,6 +306,31 @@ pub(super) fn resolve_function_property<'a, V: AsVertex<Vertex<'a>> + 'a>(
     }
 }
 
+pub(super) fn resolve_method_receiver_property<'a, V: AsVertex<Vertex<'a>> + 'a>(
+    contexts: ContextIterator<'a, V>,
+    property_name: &str,
+) -> ContextOutcomeIterator<'a, V, FieldValue> {
+    match property_name {
+        "by_value" => resolve_property_with(contexts, |vertex| {
+            let receiver = vertex.as_method_receiver().expect("not a MethodReceiver");
```
obi1kenobi: This is an example of why uniform naming matters: `MethodReceiver` here is referencing the enum variant, but what it *actually* should be referencing is a type in the Trustfall schema. The issue is that we expected a different Trustfall schema type, and it just so happens the vertex variant type didn't match either.

The distinction is important: if we expected an enum but got a struct vertex, those are both `VertexKind::Item` or `VertexKind::PositionedItem` variants but we'll still panic about it regardless.

obi1kenobi (commented):
Really nice work! This is 99% ready to merge, with just a handful of nitpicks to resolve first.

> one thing I need to dive into would be that we wrap a lot of things in a Rc, is it because of how trustfall engine works, the examples there seem to do the same, but I think we don't need this for the Receiver?

Trustfall will clone vertices often, so if a variant is expensive to clone then we wrap it in `Rc`. The new variant you added is cheap to clone so no `Rc` is needed.

src/adapter/edges.rs
Frank-III: you are right!

src/adapter/vertex.rs
Frank-III: this was a mistake! I actually have the same thought, would change it to be the same as the schema

Frank-III:
Thank you so much for the review! would fix it

test_crates/method_self_receivers/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    pub fn by_pinned_mut_ref(self: std::pin::Pin<&mut Self>) {}
+
+    pub fn by_ref_pinned_mut_ref(self: &std::pin::Pin<&'_ mut Self>) {}
+
+    pub fn by_ref_pinned_mut_ref_lifetime<'a>(self: &std::pin::Pin<&'a mut Self>) {}
+
+    pub fn by_mut_ref_pinned_mut_ref(self: &mut std::pin::Pin<&mut Self>) {}
+
+    pub fn by_mut_ref_pinned_mut_ref_lifetime(self: &mut std::pin::Pin<&'_ mut Self>) {}
```
Frank-III, line 46: I used https://grep.app/ to  find more examples, and it seems there are some usage of `Pin<&'_ mut Self>`, `Pin<&'a mut Self>` so I added them to the test crate

src/adapter/receiver.rs
```diff
@@ -1,10 +1,11 @@
 use rustdoc_types::{GenericArgs, Type};
+use std::borrow::Cow;
```
obi1kenobi, line 1: nit (and I'll eventually figure out how to get rustfmt to do this) but for readability it's usually a good idea to have built-in imports (`std / core` etc.) at the top then 3rd party imports afterward.

src/adapter/receiver.rs
```diff
@@ -68,25 +69,36 @@ fn extract_kind_string(ty: &Type) -> String {
...
+                                }
+                                _ => unreachable!("should not encounter infer"),
                             })
                             .collect();
-                        format!("{}<{}>", name, args_str.join(", "))
+
+                        let args_joined = args_str
+                            .iter()
+                            .map(|cow| cow.as_ref())
+                            .collect::<Vec<_>>()
```
obi1kenobi, line 94: Do we have to `.collect()` first to join, or can we join directly without allocating first?

src/adapter/properties.rs
```diff
@@ -312,20 +312,20 @@ pub(super) fn resolve_method_receiver_property<'a, V: AsVertex<Vertex<'a>> + 'a>
...
         "by_mut_reference" => resolve_property_with(contexts, |vertex| {
-            let receiver = vertex.as_method_receiver().expect("not a MethodReceiver");
+            let receiver = vertex.as_receiver().expect("not a MethodReceiver");
             receiver.by_mut_reference().into()
         }),
         "kind" => resolve_property_with(contexts, |vertex| {
-            let receiver = vertex.as_method_receiver().expect("not a MethodReceiver");
-            receiver.kind().into()
+            let receiver = vertex.as_receiver().expect("not a MethodReceiver");
+            receiver.kind().as_ref().into()
```
obi1kenobi: Almost certainly already on your radar, but just in case: we want all the `.expect()` messages to reference `Receiver` instead of `MethodReceiver`.

obi1kenobi (commented):
Excellent idea on the new test cases, and I love the initiative to proactively find more edge cases! 💯

A couple more nits and this is good to go.

src/adapter/receiver.rs
Frank-III: I see, good point!

src/adapter/receiver.rs
Frank-III: yes we need to `.collect()`

obi1kenobi (approved):
Excellent work!

PR #819, opened 2025-03-29
Pin the rustdoc version for `v36` to the latest nightly that supports v36

CLIDragon:
**Purpose of change**
When developing on a rustdoc version that is not the latest supported, or when testing changes across different rustdoc versions, the current workflow involves manually setting the nightly version. 

**Describe the solution**
Set the nightly version to the latest that supports v36. I know that on Zulip you mentioned that toolchain pinning for main was inappropriate - to be clear, I am only proposing pinning the toolchain version for branches which are known to have failures - i.e. `v36`, `v37`, and `v39`. 

If this is accepted, then I'll go ahead and make PR's for `v37` and `v39`.

**Testing**
Set the version and verified that running `regenerate_test_rustdocs.sh` followed by `cargo test` passes all tests.

**Additional Context**
I borrowed the formatting of this PR from https://github.com/CleverRaven/Cataclysm-DDA, since I find it helps me to organise my thoughts. If there's a particular format you would like me to use for my PRs in the future please let me know. 

I found the last applicable nightly version using https://github.githistory.xyz/rust-lang/rust/blob/master/src/rustdoc-json-types/lib.rs. Due to differences between git commit times and rust nightly publish dates, it's possible that the chosen nightly isn't actually the latest by a few days, so I manually verified that picking a later date caused failure.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -2,4 +2,5 @@
 # emits the desired format.
 
 [toolchain]
-channel = "nightly"
+# Pin to the latest nightly version that supports version 36.
+channel = "nightly-2024-12-02"


CLIDragon:
Apologies for the slightly messed up GitHub timeline. I hadn't realised that specifying the PR number in the commit message would cause GitHub to preserve references to removed commits.

rust-toolchain.toml
```diff
@@ -2,4 +2,5 @@
 # emits the desired format.
 
 [toolchain]
-channel = "nightly"
+# Pin to the latest nightly version that supports version 36.
+channel = "nightly-2024-12-02"
```
obi1kenobi, line 6: We could also just pin to stable like the CI script does: https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/7d7574a62d03f679a8c90c7738a28112bdf4935b/.github/workflows/ci.yml#L69

Is there a benefit to nightly specifically? For these old branches, it's all the same, but it's a bit easier to not have to look up what the latest nightly in a given rustdoc version is as a new version comes out.

obi1kenobi (commented):
Huh, I'd forgotten we have a toolchain file. I wonder why we added it 😅

obi1kenobi:
The PR description format is good! Thanks for trying it out.

And I don't really care about the PR timeline, so don't worry about it.

rust-toolchain.toml
CLIDragon: The main reason to pin nightly was really to support #820 and `scripts/regenerate_test_rustdocs.sh`. See #820 for a more detailed discussion, but the main gist is that pinning makes re-running tests for an older version of nightly a significantly easier process.

obi1kenobi: > pinning makes re-running tests for an older version of nightly

This is what I'm confused by: why are you re-running tests on anything other than the main development branch? I don't quite understand the motivation behind it, and I'm worried about you potentially wasting your time on something bigger.

CLIDragon: > This is what I'm confused by: why are you re-running tests on anything other than the main development branch? I don't quite understand the motivation behind it, and I'm worried about you potentially wasting your time on something bigger.

See https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/820#issuecomment-2764371984 for an explanation of why I think this change is reasonable. My comments were a bit out of order; I wasn't expecting such a fast reply 😅. In my particular case, I needed to pin to `v39` and this solution was more robust than doing it on an individual scale using `rustup override set`.

obi1kenobi: Please don't do local development on v39. Instead, use v43 with the latest nightly.

I cannot stress this enough: we *do not care* about v39-specific improvements. It is a historical version that we will drop entirely in a few months. We have no room in the maintenance budget for it.

All improvements must target the v43 branch and will be evaluated only on the basis of their impact on the v43 branch. They will then be backported to the best of our ability to older supported branches when possible. If those older branches get a benefit out of that — great! If the benefit isn't as big as it could have been if the feature were built differently — too bad! In a few months, v43 will be the tail end of what we support and the leading version will probably be v48, so that benefit will become available across all supported rustdoc JSON versions anyway.

CLIDragon:
From https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/820#issuecomment-2764578049:
> I'm sorry about my extreme stance on this, but I don't want to have to begin doing extra bookkeeping here. 

I'm interpreting this to mean that this PR has no chance of being merged, so I am going to close it.

PR #820, opened 2025-03-29
Fix in place the cargo toolchain in `prepare_benchmark_data.sh`

CLIDragon:
**Purpose of change**
Explicitly setting the toolchain to `+nightly` overrides the toolchain set in rust-toolchain.toml. In practice, this means that the script has to be hand-edited with the nightly version that corresponds to the version of `trustfall-rustdoc-adapter` that is being worked on.

**Describe the solution**
Set the nightly version to the same version that is run in the source repository. If a specific version is pinned in rust-toolchain.toml, this will use that version, otherwise it will use the default system nightly version.

I don't think that the string manipulation is the best way to extract the toolchain version, but I am very unfamiliar with bash - any guidance would be appreciated.

### DIFF
--- a/scripts/prepare_benchmark_data.sh
+++ b/scripts/prepare_benchmark_data.sh
@@ -46,9 +46,19 @@ else
 	tar -x -f "${source_tarball}" -C "${cache_path}"
 fi
 
+cd "${repo_path}" || bail "failed to cd into the source repository"
+
+# Cargo outputs a version string that looks like:
+# cargo 1.85.0-nightly (c86f4b3a1 2024-12-24)
+# In order to get the date from the end, we get all characters after the
+# last ' ' and then remove the final parentheses.
+cargo_version=`cargo --version`
+cargo_version="${cargo_version##* }"
+cargo_version="${cargo_version%?}"
+
 cd "${source_dir}/sdk/ec2" || bail "failed to cd into the extracted sources"
 
 echo "Generating rustdoc JSON..."
-RUSTDOCFLAGS="-Z unstable-options --document-private-items --document-hidden-items --output-format=json --cap-lints=allow" cargo +nightly doc --lib --no-deps || bail "failed to generate rustdoc JSON"
+RUSTDOCFLAGS="-Z unstable-options --document-private-items --document-hidden-items --output-format=json --cap-lints=allow" cargo +nightly-${cargo_version} doc --lib --no-deps || bail "failed to generate rustdoc JSON"
 
 cp -v "$source_dir/target/doc/aws_sdk_ec2.json" "$bench_data_path"


obi1kenobi:
> **Purpose of change** Explicitly setting the toolchain to `+nightly` overrides the toolchain set in rust-toolchain.toml. In practice, this means that the script has to be hand-edited with the nightly version that corresponds to the version of `trustfall-rustdoc-adapter` that is being worked on.

We only benchmark and optimize on the latest version, then port whatever we can to older versions. Versions move forward super quickly (more than once per Rust release on average) so it isn't feasible to do anything else.

In practice, except for very rare periods of time, `+nightly` is the right thing to be using for optimizations. The recent week-and-a-half period is the only outlier of anywhere close to that magnitude we've had in three years of this repo existing, and I don't foresee another outlier like it — it was caused by a change in how rustc does attribute parsing that was seismic in size, the likes of which will probably not happen again for years to come. We normally have a branch for new rustdoc JSON versions within a day or two.

To be clear, I'm not opposed to this, but I'm skeptical whether it's more likely to be useful or tech debt in the long run:
- It's useful every time `+nightly` wouldn't have been the right thing for the script to do, which we expect will be very, very few times.
- It's tech debt anytime someone has to figure out what the complex bash incantation does, tweak it to do something else, or debug something going wrong anywhere in its vicinity.

How often do these two events happen relative to each other? Picking between choices "first happens way more" / "about the same" / "second happens way more", I'd pick "about the same". How about you?

CLIDragon:
**Scenario A**
There's a bug in an old version or a fix is being backported. In order to work on the old version, you require a particular nightly rust toolchain.

**Scenario B**
A new contributor checks out the repository, aiming to work on a particular version. 
1. They're working on an older version. They have to go manually lookup the latest version.
2. They're working on the current version, but their cargo is not the latest. They have to install the latest nightly (is there a way to set the project to always track the latest nightly? Or pin a minimum nightly?)
3. They're working on the current version, but they got unlucky and the current version isn't the latest version of rustdoc.  Now they have to go lookup the latest version that supports rustdoc. 

### #819 - Toolchain Pinning
While writing this, I realised that really the main argument is in favor of #819 and `regenerate_test_rustdocs.sh`. In scenario A, B(1), or B(3), the contributor has to lookup the latest nightly version that is supported. Given that you will generally know the latest supported version at the time that you add support for the next version, pinning the toolchain will be fairly easy at that time, and save effort later on.

### Benchmarking 
> How often do these two events happen relative to each other? Picking between choices "first happens way more" / "about the same" / "second happens way more", I'd pick "about the same". How about you?

If new versions are supported as rapidly as you say, then I agree that the first scenario is quite rare. Especially as the typical contributor is unlikely to be regularly updating to the latest nightly rust. Regarding the second, I would argue the bash script has already reached the point of being technical debt, and a few extra lines of bash won't make it worse. 

I would be happy to replace the whole thing with a Python script.

scripts/prepare_benchmark_data.sh
```diff
@@ -46,9 +46,19 @@ else
...
+# last ' ' and then remove the final parentheses.
+cargo_version=`cargo --version`
+cargo_version="${cargo_version##* }"
+cargo_version="${cargo_version%?}"
+
 cd "${source_dir}/sdk/ec2" || bail "failed to cd into the extracted sources"
 
 echo "Generating rustdoc JSON..."
-RUSTDOCFLAGS="-Z unstable-options --document-private-items --document-hidden-items --output-format=json --cap-lints=allow" cargo +nightly doc --lib --no-deps || bail "failed to generate rustdoc JSON"
+RUSTDOCFLAGS="-Z unstable-options --document-private-items --document-hidden-items --output-format=json --cap-lints=allow" cargo +nightly-${cargo_version} doc --lib --no-deps || bail "failed to generate rustdoc JSON"
```
obi1kenobi, line 62: If I have done `rustup install nightly` and not `rustup install nightly-2025-03-18`, but my installed nightly does happen to be the right date's nightly, can you double-check that this invocation will actually reuse my local install and not pull in a second copy of the toolchain?

I'm actually not sure that will happen, because I don't think `rustup` is that smart. I think I recently had `stable` installed (currently `1.85.1`) and requested `+1.85.1`, and ended up downloading a second copy. We don't want to cause that behavior.

obi1kenobi:
Out of curiosity, have you already observed that CI on older versions runs on either beta or stable Rust, not nightly? This means we don't actually keep track of which historical nightly versions use which rustdoc JSON format.

Scenario A isn't realistic. We wouldn't use nightly on anything other than the latest branch, because using it is an anti-feature: `cargo-semver-checks` makes promises about things working on a set of *stable and beta Rust versions*, so that's what we use for development and testing. We only use nightly as the "about-to-be-released beta version," that's it. There's no stability or support guarantee for current or historical nightly versions outside of that.

Scenario B also isn't something we support either. If the contributor was intending to contribute to `cargo-semver-checks`, their contribution should be against the latest branch — otherwise the backporting automation won't work properly. If they were intending to do work for their own use cases, that's a fork that isn't currently intending to merge back into upstream and they are of course welcome to any changes they'd like to do.

> Given that you will generally know the latest supported version at the time that you add support for the next version, pinning the toolchain will be fairly easy at that time, and save effort later on.

I don't track the nightly version at which rustdoc JSON switched formats. Even after a PR is accepted, the rust-lang/rust merge queue makes it tricky to know when it'll get merged, and that's in turn tricky to figure out relative to the cutoff for the nightly build.

I'm sorry about my extreme stance on this, but I don't want to have to begin doing extra bookkeeping here. `cargo-semver-checks` is a project that's always on the edge of infeasibility due to technical complexity and maintenance requirements, so I'm *extremely* wary of increasing those requirements even by seemingly trivial amounts. Maintainability concerns killed multiple prior attempts at building a SemVer linter for Rust, and I don't want `cargo-semver-checks` to be their next victim.

obi1kenobi:
Ah, I think I figured out the problem. Taking a step back.

`RUSTC_BOOTSTRAP` is a magic env variable that lets us generate rustdoc JSON from a non-nightly Rust release. It also does a bunch of other things, making it a very dangerous tool in general, but in the limited context we use it in, it's safe — although it doesn't make the Rust team particularly happy that we're using it.

I want to emphasize that, for reasons of maintenance burden and maintainer bandwidth, *we are not interested* in benchmarking on anything other than the leading supported version (v43 currently), so the benchmark script should always use nightly. But if we had wanted to benchmark an older version (and we don't!), we'd use a stable or beta Rust build and that env var.

CLIDragon:
I pin my version of nightly for this project using `rust-toolchain.toml` (and I'm not saying that this project should always pin to a version of nightly - the only project pinning I think should happen should be as laid out in #819). I used to pin to `v39`. Once `v43` landed a few days ago, I pinned to that, and presumably once `v44` lands I will pin to it. I use `rust-toolchain.toml` so I only have to change my nightly-rust version for `trustfall-rustdoc-adapter`. Everywhere else on my system remains unaffected. 

What this script change does is use the version of rust-nightly that has been set in the project directory. Because I have pinned using `rust-toolchain.toml` it means that if I regenerate the benchmark data (say, because the latest supported version changed from `v39` to `v43` or `v43` to `v44`), it will generate with the correct version of rust-nightly. It's not okay to just use whatever my system nightly version is, because that could be `v43` or it could be `v24`  or anywhere in between, and I don't expect `trustfall-rustdoc-adapter` to prescribe the version of nightly for my whole system.

I propose adding the script change to the repository, so that if, in the future, another contributor wants to have two different nightly versions on their system, they wouldn't have to modify the script or hand-generate the correct benchmark data. I don't believe that the modifications add technical debt, because it doesn't add any failure modes that weren't present in the original script. If any modifications at all add technical debt, then I think it would be best to replace the entire script to deal with it now.

Working with non-latest versions of `rustdoc-types` is incidental, but a non-goal of this change.

scripts/prepare_benchmark_data.sh
CLIDragon: As far as I can tell, this is intentional `rustup` behavior. Your current version of `stable` or `nightly` is transient, and will be replaced whenever you update. By downloading a second version it ensures that even if you `rustup +stable update` `+1.85.1` will always continue to work. 

> We don't want to cause that behavior.

By using nightly rust, we already do. If I have a project that depends on `204-11-07` (for instance) and your project that (currently) depends on `>= 2025-03-06` then I have to have two versions on my system. In general, for `n` projects that depend on nightly you have at least `n` nightly versions. I don't believe one more makes a difference.

If this ends up being the only blocker, then fixing it will simply mean checking that the version of rust invoked inside the `ec2` folder is at least as old as the version as the one in the crate. This will mean the script fails when benchmarking on an older version (explicitly not supported) or if your nightly version is too new (rare), both of which are acceptable.

obi1kenobi: I'm saying something else: that this project by itself will pull in two separate copies of nightly with this PR.

The toolchain file will make the initial `cargo` call in the script behave as if it's `cargo +nightly`. This will install the latest nightly toolchain.

Then this script will look up the date in the toolchain and add it to the explicit `+` version request. Then `rustup` will download a second copy of the toolchain. This is undesirable — we have lots of contributors all over the world not all of whom have a very high-bandwidth internet connection.

CLIDragon: That doesn't match with my experience. I've never found `cargo +nightly` to install the latest nightly toolchain. It just uses whatever toolchain on the system is currently set as `nightly`, whether that be the latest of 6 months out of date.

obi1kenobi: I agree! But note that the proposed change takes the `cargo +nightly` invocation and replaces it with `cargo +nightly-${cargo_version}` AKA something like `cargo +nightly-2025-03-31`, which `rustup` will treat as a version separate from `+nightly`. Hence my concern :)

PR #822, opened 2025-04-14
Add `maybe_sized` property to `GenericTypeParameter`

KhaledSayed04:
*(PR description by @obi1kenobi to match the maintainer edits to this PR that were made pre-merge)*

Add a `maybe_sized: Boolean!` property to `GenericTypeParameter`.

The computation of the property does only local reasoning about bounds. In particular:
- `T: ?Sized + Sized` is considered sized.
- `T: ?Sized + SomeTrait` is considered maybe-sized, even if `SomeTrait: Sized`.

Implementing correct global reasoning for precise `?Sized` determinations requires either cross-crate querying, or explicit information on this in rustdoc JSON.

### DIFF
--- a/test_crates/generic_type_param_maybe_sized/src/lib.rs
+++ b/test_crates/generic_type_param_maybe_sized/src/lib.rs
@@ -49,7 +49,7 @@ pub fn generic_fn2<T: core::marker::Sized + ?core::marker::Sized>(value: &T) {}
 pub fn generic_fn3<T: Sized + ?core::marker::Sized>(value: T) {}
 
 // `T` is sized. `SizedRenamed` is an import rename of the built-in `Sized` trait.
-pub fn generic_f3a<T: SizedRenamed + ?core::marker::Sized>(value: T) {}
+pub fn generic_fn3a<T: SizedRenamed + ?core::marker::Sized>(value: T) {}
 
 // `T` is sized. The `?Sized` is overridden by both explicit bounds: the `where` and the `:` bound.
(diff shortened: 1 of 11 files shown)


obi1kenobi:
Hey! Just checking: are you still interested and able to move this PR forward?

I appreciate your efforts so far, and if you're busy with other things, I'm happy to continue and drive it to merge from here.

test_crates/generic_type_param_maybe_sized/src/lib.rs
```diff
@@ -0,0 +1,43 @@
+use std::boxed::Box;
+pub struct GenericStruct<T: ?core::marker::Sized, U: core::marker::Sized, V> {
```
Frank-III, line 13: would you like to add some tests to include traits like `UserSafeSized`(appeared in the `rust` repo) that ends with `Sized`, I believe your code works since trait-bound modifier only work for `Sized` trait

obi1kenobi:
I see this is marked ready for review, but the code seems quite messy, triggers a number of lints, and has some of the same bugs we discussed in the past. I'm assuming the review request might have been a mis-click?

KhaledSayed04:
> I see this is marked ready for review, but the code seems quite messy, triggers a number of lints, and has some of the same bugs we discussed in the past. I'm assuming the review request might have been a mis-click?

I don't notice these "bugs" (except you means the time complexity). And I opened it for review to make you able to continue and drive it to merge from here if it's urgent as I'm busy nowadays.

PR #855, opened 2025-06-14
Omit trait-provided items from impl_index when overridden in impl.

obi1kenobi:
Ensure we don't get duplicate items when looking up items by name.

- **Omit trait-provided items from `impl_index` when overridden in `impl`.**
- **Rename `impl_index` to `impl_method_index` to reflect its role.**


src/adapter/optimizations/method_lookup.rs
```diff
@@ -161,6 +154,7 @@ fn resolve_impl_method_by_name<'a>(
     method_name: &str,
 ) -> VertexIterator<'a, Vertex<'a>> {
     if let Some(method_ids) = impl_index.get(&(impl_owner_id, method_name)) {
+        dbg!(method_ids);
```
obi1kenobi:
```suggestion
```

PR #902, opened 2025-07-09
Add a per-item-kind index of public items

CLIDragon:
This PR is based on #676. Due to merge conflicts it was easier to create a new PR than attempt to build on that one, but all the points made in that PR remain current.

**Purpose of Change**
From @obi1kenobi in #676:
> Most items we'll see in the crate either aren't public, or aren't top-level items that would be returned by the `Crate -[item]-> Item` edge. For example, many of those items are methods, impls, trait associated types, associated constants, etc.
> 
> In this PR, we ensure we don't bother iterating over such items in queries where they can't possibly match. Queries that look at public structs should only iterate over structs, ones that look at enums should only iterate over enums, etc.
> 
> This has a substantial performance effect: it speeds up the median `cargo-semver-checks` lint by ~100x, and makes it so that a tiny handful (~5) of lints dominate all linting runtime — the combined runtime of all other lints becomes negligible. In end-to-end terms, `cargo-semver-checks` becomes about ~35% faster on our largest workload, but this number is misleadingly small w.r.t. the impact of this optimization: the new runtime is now equal to ~101% of the slowest lint, whereas that number was ~140% previously. If we can now optimize that slow lint in particular, we'll get tens of percent of additional speedup from the optimization in this PR.

On top of that, when calling `resolve_crate_items` and resolving a path, all items matching that path are returned from `imports_index`. However, a single path might have dozens of imports, and only some of these items will be of the right type (e.g. a Trait or Enum). 

**Describe the Solution**
In order to return only the items of the right type, we can take the intersection of `imports_index` and `pub_item_kind_index`. However, as both of these were vectors, it was very inefficient to find their intersection. Replacing the vectors inside `pub_item_kind_index` with maps allows efficiently taking the intersection of indices.

As noted in https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/676#issuecomment-3051767440, there is a decent improvement in performance thanks to this change (roughly a 40% improvement in total time spent in queries). Additionally, since the indices are generally quite small there is a negligible effect on indexed crate creation time.

**Describe alternatives you've considered**
- Use an `Rc<str>` instead of `Arc<str>`. I think it would be safe, but I don't know enough about trustfall's internals to be sure.
- ~~Inline `PubItemKindIndex::contains` into `resolve_items_by_importable_path_field_value`~~. This was done.
- Create the index in parallel. Benchmarks showed parallel construction used in the original commit was significantly slower than the single-threaded version, taking over a second compared to the single-threaded version resolving in milliseconds.

### DIFF
--- a/src/adapter/optimizations/item_lookup.rs
+++ b/src/adapter/optimizations/item_lookup.rs
@@ -8,6 +8,7 @@ use trustfall::{
 };
 
 use super::super::{RustdocAdapter, origin::Origin, vertex::Vertex};
+use std::sync::Arc;
 
 use crate::IndexedCrate;
 
@@ -31,16 +32,26 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
         // statically vs dynamically, so we check the dynamic case first since
         // it might be more specific.
         if let Some(dynamic_value) = neighbor_info.dynamically_required_property("path") {
-            return dynamic_value.resolve_with(&adapter, contexts, |vertex, candidate| {
+            return dynamic_value.resolve_with(&adapter, contexts, move |vertex, candidate| {
                 let crate_vertex = vertex.as_indexed_crate().expect("vertex was not a Crate");
                 let origin = vertex.origin;
-                resolve_items_by_importable_path(crate_vertex, origin, candidate)
+                resolve_items_by_importable_path(
+                    crate_vertex,
+                    origin,
+                    destination.coerced_to_type().cloned(),
+                    candidate,
+                )
             });
         } else if let Some(path_value) = neighbor_info.statically_required_property("path") {
             return resolve_neighbors_with(contexts, move |vertex| {
                 let crate_vertex = vertex.as_indexed_crate().expect("vertex was not a Crate");
                 let origin = vertex.origin;
-                resolve_items_by_importable_path(crate_vertex, origin, path_value.clone())
+                resolve_items_by_importable_path(
+                    crate_vertex,
+                    origin,
+                    destination.coerced_to_type().cloned(),
+                    path_value.clone(),
+                )
             });
         }
     }
@@ -83,10 +94,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .free_functions
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "Struct" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -97,10 +108,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .structs
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "Enum" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -111,10 +122,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .enums
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "Union" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -125,10 +136,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .unions
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "Trait" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -139,10 +150,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .traits
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "ImplOwner" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -153,12 +164,12 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .structs
-                                .iter()
-                                .chain(crate_vertex.pub_item_kind_index.enums.iter())
-                                .chain(crate_vertex.pub_item_kind_index.unions.iter())
+                                .values()
+                                .chain(crate_vertex.pub_item_kind_index.enums.values())
+                                .chain(crate_vertex.pub_item_kind_index.unions.values())
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "Constant" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -169,10 +180,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .free_consts
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "Static" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -183,10 +194,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .statics
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "GlobalValue" => {
                     // const or static
@@ -198,8 +209,8 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .free_consts
-                                .iter()
-                                .chain(crate_vertex.pub_item_kind_index.statics.iter())
+                                .values()
+                                .chain(crate_vertex.pub_item_kind_index.statics.values())
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
                     });
@@ -213,10 +224,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .decl_macros
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "ProcMacro"
                 | "FunctionLikeProcMacro"
@@ -230,10 +241,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .proc_macros
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 "Module" => {
                     return resolve_neighbors_with(contexts, move |vertex| {
@@ -244,10 +255,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .modules
-                                .iter()
+                                .values()
                                 .map(move |item| origin.make_item_vertex(item)),
                         )
-                    })
+                    });
                 }
                 _ => {}
             }
@@ -264,15 +275,24 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
 fn resolve_items_by_importable_path<'a>(
     crate_vertex: &'a IndexedCrate,
     origin: Origin,
+    destination_type: Option<Arc<str>>,
     importable_path: CandidateValue<FieldValue>,
 ) -> VertexIterator<'a, Vertex<'a>> {
     match importable_path {
         CandidateValue::Impossible => Box::new(std::iter::empty()),
-        CandidateValue::Single(value) => {
-            resolve_items_by_importable_path_field_value(crate_vertex, origin, &value)
-        }
+        CandidateValue::Single(value) => resolve_items_by_importable_path_field_value(
+            crate_vertex,
+            origin,
+            destination_type,
+            &value,
+        ),
         CandidateValue::Multiple(values) => Box::new(values.into_iter().flat_map(move |value| {
-            resolve_items_by_importable_path_field_value(crate_vertex, origin, &value)
+            resolve_items_by_importable_path_field_value(
+                crate_vertex,
+                origin,
+                destination_type.clone(),
+                &value,
+            )
         })),
         _ => {
             // fall through to slow path
@@ -284,6 +304,7 @@ fn resolve_items_by_importable_path<'a>(
 fn resolve_items_by_importable_path_field_value<'a>(
     crate_vertex: &'a IndexedCrate,
     origin: Origin,
+    destination_type: Option<Arc<str>>,
     value: &FieldValue,
 ) -> VertexIterator<'a, Vertex<'a>> {
     let path_components: Vec<&str> = value
@@ -298,7 +319,18 @@ fn resolve_items_by_importable_path_field_value<'a>(
         .expect("crate's imports_index was never constructed")
         .get(path_components.as_slice())
     {
-        resolve_item_vertices(origin, items.iter().map(|(item, _)| item).copied())
+        resolve_item_vertices(
+            origin,
+            items
+                .iter()
+                .map(|(item, _)| item)
+                .copied()
+                .filter(move |item| {
+                    crate_vertex
+                        .pub_item_kind_index
+                        .contains(destination_type.clone(), item.id)
+                }),
+        )
     } else {
         // No such items found.
         Box::new(std::iter::empty())
--- a/src/adapter/optimizations/item_lookup.rs
+++ b/src/adapter/optimizations/item_lookup.rs
@@ -288,14 +288,14 @@ fn resolve_items_by_importable_path<'a>(
         CandidateValue::Single(value) => resolve_items_by_importable_path_field_value(
             crate_vertex,
             origin,
-            destination_type,
+            destination_type.as_deref(),
             &value,
         ),
         CandidateValue::Multiple(values) => Box::new(values.into_iter().flat_map(move |value| {
             resolve_items_by_importable_path_field_value(
                 crate_vertex,
                 origin,
-                destination_type.clone(),
+                destination_type.as_deref(),
                 &value,
             )
         })),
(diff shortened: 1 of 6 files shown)


src/indexed_crate.rs
```diff
@@ -231,154 +231,139 @@ pub struct IndexedCrate<'a> {
...
+            free_functions: HashMap::with_capacity_and_hasher(capacity, rustc_hash::FxBuildHasher),
+            structs: HashMap::with_capacity_and_hasher(capacity, rustc_hash::FxBuildHasher),
+            enums: HashMap::with_capacity_and_hasher(capacity, rustc_hash::FxBuildHasher),
+            traits: HashMap::with_capacity_and_hasher(capacity, rustc_hash::FxBuildHasher),
+            unions: HashMap::with_hasher(rustc_hash::FxBuildHasher),
+            modules: HashMap::with_capacity_and_hasher(64, rustc_hash::FxBuildHasher),
+            statics: HashMap::with_hasher(rustc_hash::FxBuildHasher),
+            free_consts: HashMap::with_hasher(rustc_hash::FxBuildHasher),
+            decl_macros: HashMap::with_hasher(rustc_hash::FxBuildHasher),
+            proc_macros: HashMap::with_hasher(rustc_hash::FxBuildHasher),
```
obi1kenobi: After we merge the most recent changes on the main branch, we should be able to import `HashMap` from our `hashtables.rs` which should take care of the feature-flagging and should mean we don't have to duplicate this method.

src/indexed_crate.rs
```diff
@@ -231,154 +231,139 @@ pub struct IndexedCrate<'a> {
...
-        #[cfg(not(feature = "rayon"))]
         let iter = crate_.index.values();
+        let init = PubItemKindIndex::with_capacity_hint(crate_.index.len());
 
-        #[cfg(feature = "rayon")]
-        let init = || Self::with_capacity_hint(crate_.index.len());
-        #[cfg(not(feature = "rayon"))]
-        let init = Self::with_capacity_hint(crate_.index.len());
-
-        let folded = iter.fold(init, |mut acc, item| {
```
obi1kenobi: You mentioned:
> The parallel construction used in the original commit was significantly slower than the single-threaded version so I removed it.

I'm curious:
- What crates did you test this on?
- What were the raw results before and after?
- How many runs did you do, and what was the uncertainty range in each?
- What hardware did you run it on? (CPU architecture, number of cores, etc.)

Historically, we added parallel index construction because index construction took a long time on very large crates. Adding a tiny amount of overhead to a small crate is generally worth it if we cut a large amount of time from very large crates.

If the outcomes are not conclusive ("everyone is better off") or if you didn't have the time to run a detailed-enough analysis like this, then let's not make this change here and instead save it for a future PR.

src/indexed_crate.rs
```diff
@@ -231,154 +231,139 @@ pub struct IndexedCrate<'a> {
...
-                value.modules.extend(c.modules);
-                value.statics.extend(c.statics);
-                value.free_consts.extend(c.free_consts);
-                value.decl_macros.extend(c.decl_macros);
-                value.proc_macros.extend(c.proc_macros);
+    /// Returns true if the index corresponding to the type given by
+    /// `destination_type` contains an element with the id `item_id`.
+    /// Conservatively, returns true if the destination is None or of a type
+    /// that isn't contained in the indexes.
+    pub fn contains(&self, destination_type: Option<Arc<str>>, item_id: Id) -> bool {
```
obi1kenobi: We don't use the fact that we get an `Arc` so let's not demand it here. Otherwise we're tying up the caller's ability to change implementations needlessly.
```suggestion
    pub fn contains(&self, destination_type: Option<&str>, item_id: Id) -> bool {
```

src/adapter/optimizations/item_lookup.rs
```diff
@@ -83,10 +94,10 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
                             crate_vertex
                                 .pub_item_kind_index
                                 .free_functions
-                                .iter()
+                                .values()
```
obi1kenobi, line 98: All the `.values()` calls here are on hashtables, which have non-deterministic iteration order. This is something we want to avoid, because it makes debugging hard: outcomes are non-reproducible.

Because `cargo-semver-checks` is an OSS project, its priorities are a bit different than corporate software. We're very "understaffed" compared to an equivalent corporate project  with similarly wide use, so maintenance issues (issue triage, debugging, bugfixing etc.) are very costly. We have to prioritize maintainability above performance, or else the project will stall and die.

One option is to use `BTreeMap` instead of hashmap indexes. Another option is to come up with a different index layout. What do you think?

src/indexed_crate.rs
```diff
@@ -231,154 +231,139 @@ pub struct IndexedCrate<'a> {
...
+                _ => true,
             }
-
-            value
-        };
-
-        folded
+        } else {
+            true
+        }
```
obi1kenobi: nit: The "if `None` return `true`" side seems not particularly worth it here. Consider refactoring this to push the `None` check into the caller, and have `contains()` only take a `&str`.

src/adapter/optimizations/item_lookup.rs
```diff
@@ -298,7 +319,18 @@ fn resolve_items_by_importable_path_field_value<'a>(
...
+        resolve_item_vertices(
+            origin,
+            items
+                .iter()
+                .map(|(item, _)| item)
+                .copied()
+                .filter(move |item| {
+                    crate_vertex
+                        .pub_item_kind_index
+                        .contains(destination_type.clone(), item.id)
```
obi1kenobi: See if we can avoid cloning the `Arc` on every iteration? Based on a quick glance, it feels like we should be able to move it into the closure without cloning on every call.

src/adapter/optimizations/item_lookup.rs
```diff
@@ -264,15 +275,24 @@ pub(crate) fn resolve_crate_items<'a, V: AsVertex<Vertex<'a>> + 'a>(
 fn resolve_items_by_importable_path<'a>(
     crate_vertex: &'a IndexedCrate,
     origin: Origin,
+    destination_type: Option<Arc<str>>,
```
obi1kenobi, line 283: This is a good addition, but now the function's contract is dissonant with its name: we're doing a lot more than just resolving items by importable path here. We're probably at the limit of how big the function's name can get, though.

Please add doc comments here and in the other function explaining the new post-filtering behavior.

src/adapter/optimizations/item_lookup.rs
```diff
@@ -8,6 +8,7 @@ use trustfall::{
 };
 
 use super::super::{RustdocAdapter, origin::Origin, vertex::Vertex};
+use std::sync::Arc;
```
obi1kenobi: nit: let's put this in the upper "external imports" block, not between `super` and `crate`

obi1kenobi (commented):
This is a very nice prototype! Well done!

It'll still need a bit of polishing to get it to a release-ready stage, but it'll be a massive improvement and I'm excited to help you get it there.

CLIDragon:
Benchmarking of parallel construction was done using `cargo bench --features rustc-hash,rayon`. While there is quite a lot of noise in the measurements, they are quite clear that parallel construction is significantly slower. My CPU is a Ryzen 9 7950X (16 Cores), so I don't think there is an issue with lacking core count. 
```
Benchmarking IndexedCrate/new(aws-sdk-ec2): Collecting 10 samples in estimated 5IndexedCrate/new(aws-sdk-ec2)
                        time:   [263.48 ms 276.65 ms 293.67 ms]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
Benchmarking IndexedCrate/new-parallel(aws-sdk-ec2): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target
time to 17.4s.
Benchmarking IndexedCrate/new-parallel(aws-sdk-ec2): Collecting 10 samples in esIndexedCrate/new-parallel(aws-sdk-ec2)
                        time:   [1.5417 s 1.6722 s 1.8032 s]
```
```
Benchmarking IndexedCrate/new(aws-sdk-ec2): Collecting 10 samples in estimated 5IndexedCrate/new(aws-sdk-ec2)
                        time:   [251.66 ms 253.36 ms 255.22 ms]
                        change: [-13.812% -8.4184% -3.8197%] (p = 0.01 < 0.05)
                        Performance has improved.
Benchmarking IndexedCrate/new-parallel(aws-sdk-ec2): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target
time to 20.0s.
Benchmarking IndexedCrate/new-parallel(aws-sdk-ec2): Collecting 10 samples in esIndexedCrate/new-parallel(aws-sdk-ec2)
                        time:   [2.1108 s 2.2609 s 2.4276 s]
                        change: [+22.046% +35.206% +51.141%] (p = 0.00 < 0.05)
                        Performance has regressed
```
I re-ran the benchmark while limiting rayon to two threads (`RAYON_NUM_THREADS = 2`), and results are much closer:
```
Benchmarking IndexedCrate/new(aws-sdk-ec2): Collecting 10 samples in estimated 9IndexedCrate/new(aws-sdk-ec2)
                        time:   [450.06 ms 450.83 ms 451.66 ms]
                        change: [+76.572% +77.942% +79.179%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking IndexedCrate/new-parallel(aws-sdk-ec2): Collecting 10 samples in esIndexedCrate/new-parallel(aws-sdk-ec2)
                        time:   [466.56 ms 472.49 ms 478.64 ms]
                        change: [-80.545% -79.102% -77.573%] (p = 0.00 < 0.05)
                        Performance has improved.
```

obi1kenobi:
Neat, nice find! Serial construction it is then — I'd just add a comment next to it that links to your comment in this PR, so that we don't forget about this and attempt to "optimize" it again in the future.

src/adapter/optimizations/item_lookup.rs
CLIDragon: We could use an `IndexMap` instead of a `HashMap`, as it preserves insertion order, same as a `Vec` would. We already have `indexmap` as a dependency thanks to `toml_edit`, so it shouldn't affect compilation time.

obi1kenobi: Sounds good, let's give it a shot. Can we use the `Fx` hasher for it as well?

Consider updating `hashtables.rs` to provide appropriate crate-level re-exports of appropriate indexmap types.

src/indexed_crate.rs
CLIDragon: Unfortunately, `with_capacity` and `new` are not compatible with using a different hash builder, due to the way they are hardcoded inside `std`. Aside from using the feature flag, the main options (as discussed [here](https://internals.rust-lang.org/t/hashmap-set-new-with-capacity-and-buildhasher/15622)) are to implement an extension trait, e.g. 
```rust
trait HashMapExt {
    fn new() -> Self;
    fn with_capacity(x: usize) -> Self;
}

impl<K, V, S> HashMapExt for HashMap<K, V, S>
where
    K: Hash + Eq,
    S: BuildHasher + Default,
{
    fn new() -> Self {
        HashMap::with_hasher(S::default())
    }

    fn with_capacity(capacity: usize) -> Self {
        HashMap::with_capacity_and_hasher(capacity, S::default())
    }
}
```
or to implement a newtype on top of `HashMap` (as in https://github.com/tkaitchuck/aHash/pull/23)

obi1kenobi: Yeah, that's fine — we can use the extension trait trick, just leave a link next to it so we know why "Chesterton's fence" is there and so we have a better idea of when we might be able to remove it (one day).

src/adapter/optimizations/item_lookup.rs
```diff
@@ -319,18 +320,22 @@ fn resolve_items_by_importable_path_field_value<'a>(
...
+            resolve_item_vertices(
+                origin,
+                items
+                    .iter()
+                    .map(|(item, _)| item)
+                    .copied()
+                    .filter(move |item| {
+                        crate_vertex
+                            .pub_item_kind_index
+                            .contains(destination_type.as_ref(), item.id)
```
obi1kenobi: Looking at this more closely, there's an opportunity to do a [loop-invariant code motion](https://en.wikipedia.org/wiki/Loop-invariant_code_motion) optimization here.

Note that this is run in a loop:
```rust
                        crate_vertex
                            .pub_item_kind_index
                            .contains(destination_type.as_ref(), item.id)
```
and if we imagine inlining that `contains()` function here, we get a loop over `match destination_type.as_ref() { ... }`.

That pattern match resolves the same way in each iteration of the loop, because `destination_type` doesn't change inside the loop. So with that in mind, let's try applying the optimization:
- pull the `match` out to just inside `if let Some(destination_type) = destination_type {`
- look up the appropriate item kind index component inside `crate_vertex.pub_item_kind_index`
- have the looped `.filter()` look up directly into that index each time

For the few special-cased destination types that look up multiple indexes, we can specialize the code further.

Rust _might or might not have_ been doing this already. But let's make sure :)

src/indexed_crate.rs
```diff
@@ -332,37 +335,31 @@ impl<'a> PubItemKindIndex<'a> {
...
+            "GlobalValue" => {
+                // const or static
+                self.free_consts.contains_key(&item_id) || self.statics.contains_key(&item_id)
+            }
+            "Macro" => self.decl_macros.contains_key(&item_id),
+            "ProcMacro" | "FunctionLikeProcMacro" | "AttributeProcMacro" | "DeriveProcMacro" => {
+                self.proc_macros.contains_key(&item_id)
+            }
+            "Module" => self.modules.contains_key(&item_id),
+            _ => true,
```
obi1kenobi: Out of curiosity, is this reachable? Right now this fails open ("the optimization doesn't apply") but I'm wondering if we're setting up a bit of an optimization trap for ourselves here — another suspicious perf behavior we'd have to debug in the future.

Perhaps time for something akin to `debug_assert!(false);` and then return `true`? Clippy will probably complain about that and then suggest a better pattern :)

CLIDragon: I swapped it for `unreachable` - after a quick search of the lints, none of them coerce an `Item` to a type that is not one of the options:
```
... on AttributeProcMacro
... on Constant
... on DeriveProcMacro
... on Enum
... on Function
... on FunctionLikeProcMacro
... on GenericConstParameter (generic_parameter)
... on GenericLifetimeParameter (generic_parameter)
... on GenericTypeParameter (generic_parameter)
... on GlobalValue
... on ImplOwner
... on Macro
... on Module
... on PlainVariant (variant)
... on ProcMacro
... on ResolvedPathType (raw_type)
... on Static
... on Struct
... on StructVariant (variant)
... on Trait
... on TupleVariant (variant)
... on Union
```

src/hashtables.rs
```diff
@@ -1,5 +1,50 @@
...
+
+#[cfg(feature = "rustc-hash")]
+pub(crate) type IndexMap<K, V> = indexmap::map::IndexMap<K, V, rustc_hash::FxBuildHasher>;
+
+#[cfg(not(feature = "rustc-hash"))]
+pub(crate) use indexmap::map::IndexMap;
+
+/// Allow using new() and with_capacity() regardless of the hash algorithm.
+/// See <https://github.com/tkaitchuck/aHash/issues/103> for more information.
+#[allow(dead_code)] // Used when rustc-hash is enabled.
```
obi1kenobi:
```suggestion
#[allow(dead_code, reason = "used when the `rustc-hash` feature is enabled")]
```

src/indexed_crate.rs
```diff
@@ -4,9 +4,11 @@ use std::{borrow::Borrow, collections::hash_map::Entry, sync::Arc};
 use rayon::prelude::*;
 use rustdoc_types::{Crate, Id, Item};
 
+#[allow(unused_imports)]
```
```suggestion
#[allow(unused_imports, reason = "used when the `rustc-hash` feature is enabled")]
```

src/indexed_crate.rs
```diff
@@ -226,6 +231,132 @@ pub struct IndexedCrate<'a> {
...
+                }
+            }
+
+            acc
+        })
+    }
+
+    /// Returns true if the index corresponding to the type given by
+    /// `destination_type` contains an element with the id `item_id`.
+    // Currently unused, but might be used in the future so it stays.
```
obi1kenobi: I'd rather pull it out of git or PR history actually. Let's remove it.

Things change in this crate quite often to the point where I have automation scripts for porting changes from one rustdoc version's branch to another. Generally speaking, those scripts work better the less code there is in the project :)

src/indexed_crate.rs
```diff
@@ -226,6 +231,132 @@ pub struct IndexedCrate<'a> {
...
+            "ProcMacro" | "FunctionLikeProcMacro" | "AttributeProcMacro" | "DeriveProcMacro" => {
+                self.proc_macros.contains_key(&item_id)
+            }
+            "Module" => self.modules.contains_key(&item_id),
+            _ => {
+                // Currently, this function is only called in the context of resolving
+                // the coercion of a crate to an item. As there are no lints that coerce
+                // an item to a type other than the ones listed above, this is unreachable.
+                // If a lint is added that does coerce to a different type, consider adding
+                // it to the index instead of making this branch reachable.
```
obi1kenobi: The "no lints use this" angle is in general not valid justification.

There's no rule that says this crate should only be usable with `cargo-semver-checks`. It also powers https://play.predr.ag/rustdoc for example. Other people may be using it too. So we can't make any _correctness_ arguments on the basis of lints, because from this crate's perspective there are no lints — only the space of valid queries arbitrary users might write. (_Performance_ arguments are a different story. But if another reasonable use case presented a perf issue solvable by a tweak in the indexing strategy, we'd certainly look into it anyway.)

If we're marking this branch unreachable in release builds, it needs to be actually provably unreachable, by analysis of the schema and the valid coercions from `Item` to any other type derived from it.

I believe that such an analysis would determine this code is actually reachable, and we'd need to `return true` here instead.

I asked to remove this function as dead code so this won't matter _here_. But the overall point is very important, so I wanted to mention it anyway.

obi1kenobi (commented):
A few last nits, otherwise should be good. I think a few of my previous comments are still open as well.

I'd recommend tweaking the PR title since it'll appear in the changelog. Let's not make people look up #676 first, it's okay to just link it from the description.

I'm excited to get this merged, it's a very nice perf improvement and hopefully just the start of a series of improvements! But no rush — just give me a ping when you've had a chance to address all the comments and I'd be happy to give it another look.

src/indexed_crate.rs
CLIDragon: I forgot about that playground actually. Part of the reason I added that was that I wasn't sure it was possible to coerce an `Item` directly to a sub-category like `TupleVariant`, but it seems like it is.

CLIDragon:
I think that's everything code-wise. I'll clean up the remaining documentation work tomorrow.

src/adapter/optimizations/item_lookup.rs
CLIDragon: I've added comments in now, but I think they're a bit short - please let me know how you think they could be improved.

CLIDragon:
@obi1kenobi I think that's everything done. I would appreciate you checking the documentation comments to see if they make sense. 

One last concern I have is that we are assuming that having an importable path is the same as having public visibility. Is this true in the general case, or does it only work due to the context of these specific functions?

obi1kenobi:
It looks good! I pushed one commit with a few more comments and one small signature change, making a function take `Option<&str>` instead of `Option<Arc<str>>` where it could get away with it. This won't have an effect on performance, but might make future refactoring easier because it now requests no more than it strictly speaking needs.

Merging this now — thanks for getting it over the finish line!

obi1kenobi (approved):

obi1kenobi:
> One last concern I have is that we are assuming that having an importable path is the same as having public visibility. Is this true in the general case, or does it only work due to the context of these specific functions?

This is true because importable path is something we compute in this crate via name analysis and reachability analysis, and it's a strictly stronger property than being `pub`.

For example:
```rust
mod priv {
    pub struct Example; // this item is not importable

    pub struct Importable;
}

pub use priv::Importable; // this re-export gives `Importable` an importable path
```

It's a compile error to `pub` re-export an item that has not declared itself `pub`. So between these properties, we know that if the `importable_path` edge is mandatory (not `@optional` and not `@fold`, unless it's `@fold @transform(op: "count") @filter(..)` that implies a count > 0) then the item must be `pub` and it must be reachable from the crate root via some chain of `pub` items (`pub mod`, `pub use` etc.).

There are many, many edge cases here. For example:
- https://predr.ag/blog/moving-and-reexporting-rust-type-can-be-major-breaking-change/
- https://predr.ag/blog/checking-semver-for-doc-hidden-items/

But our test suite is _extremely_ comprehensive in this area, so I trust our analysis quite a bit.

CLIDragon:
I didn't test `par_extend` when creating the index in parallel, and it might be worth looking into whether it produces a speedup over serial construction.

PR #903, opened 2025-07-10
Use hashtables.rs instead of cfg attributes.

CLIDragon:
While working on #902, I came across these two files which don't use `hashtables.rs`.

**Purpose of Change**
Makes code slightly cleaner.


obi1kenobi (approved):

obi1kenobi:
Nice catch, thank you!

PR #926, opened 2025-07-22
Initial work towards dynamically resolving enum variants.

CLIDragon:
**Purpose of Change**
Resolving enum variants costs a lot of time in already heavy queries. In particular, the following lints will run faster after this change.
```
enum_no_repr_variant_discriminant_changed
enum_unit_variant_changed_kind
enum_variant_marked_non_exhaustive
enum_variant_missing
partial_ord_enum_variants_reordered
```

**Describe the Change**
Implement a custom resolver for enum variants based on a new index.

**Additional Context**
If all of the above lints act only on public enums then index creation can be made faster by using `pub_item_kind_index`.

### DIFF
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeSet, num::NonZeroUsize, rc::Rc};
+use std::{collections::BTreeSet, num::NonZeroUsize};
 
 use rustdoc_types::{
     GenericBound::TraitBound, GenericParamDefKind, Id, ItemEnum, VariantKind, WherePredicate,
@@ -15,9 +15,7 @@ use crate::{
 };
 
 use super::{
-    RustdocAdapter,
-    enum_variant::LazyDiscriminants,
-    optimizations,
+    RustdocAdapter, optimizations,
     origin::Origin,
     receiver::Receiver,
     vertex::{Feature, Vertex},
@@ -449,70 +447,12 @@ pub(super) fn resolve_enum_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
     contexts: ContextIterator<'a, V>,
     edge_name: &str,
     adapter: &'a RustdocAdapter<'a>,
+    resolve_info: &ResolveEdgeInfo,
 ) -> ContextOutcomeIterator<'a, V, VertexIterator<'a, Vertex<'a>>> {
     match edge_name {
-        "variant" => resolve_neighbors_with(contexts, move |vertex| {
-            let origin = vertex.origin;
-            let enum_item = vertex.as_enum().expect("vertex was not an Enum");
-            let outer_item = vertex.as_item().expect("enum was not a vertex");
-
-            let item_index = &adapter.crate_at_origin(origin).own_crate.inner.index;
-
-            let discriminants = {
-                // Discriminants are only well-defined if either:
-                // - the enum has a defined `repr` binary representation, or
-                // - none of the enum variants contain any fields of their own.
-
-                let has_repr = outer_item.attrs.iter().any(|attr| match attr {
-                    rustdoc_types::Attribute::Repr(attr_repr) => attr_repr.int.is_some(),
-                    _ => false,
-                });
-
-                let mut has_fields_in_variants = false;
-                let variants = enum_item
-                    .variants
-                    .iter()
-                    .map(|field_id| {
-                        let inner = &item_index.get(field_id).expect("missing item").inner;
-                        match inner {
-                            ItemEnum::Variant(v) => {
-                                match &v.kind {
-                                    VariantKind::Plain => {}
-                                    VariantKind::Tuple(t) => {
-                                        has_fields_in_variants |= !t.is_empty()
-                                    }
-                                    VariantKind::Struct { fields, .. } => {
-                                        has_fields_in_variants |= fields.is_empty()
-                                    }
-                                }
-                                v
-                            }
-                            _ => unreachable!("Item {inner:?} not a Variant"),
-                        }
-                    })
-                    .collect();
-
-                if has_repr || !has_fields_in_variants {
-                    Some(Rc::new(LazyDiscriminants::new(variants)))
-                } else {
-                    None
-                }
-            };
-
-            Box::new(
-                enum_item
-                    .variants
-                    .iter()
-                    .enumerate()
-                    .map(move |(index, field_id)| {
-                        origin.make_variant_vertex(
-                            item_index.get(field_id).expect("missing item"),
-                            discriminants.clone(),
-                            index,
-                        )
-                    }),
-            )
-        }),
+        "variant" => {
+            optimizations::variant_lookup::resolve_enum_variant(adapter, contexts, resolve_info)
+        }
         _ => unreachable!("resolve_enum_edge {edge_name}"),
     }
 }
--- a/src/adapter/optimizations/variant_lookup.rs
+++ b/src/adapter/optimizations/variant_lookup.rs
@@ -133,7 +133,7 @@ pub(crate) fn resolve_enum_variant<'a, V: AsVertex<Vertex<'a>> + 'a>(
     resolve_info: &ResolveEdgeInfo,
 ) -> ContextOutcomeIterator<'a, V, VertexIterator<'a, Vertex<'a>>> {
     // If the name of the variant is required to be a particular value, we can
-    // use the variant_name_index find the corresponding variant.
+    // use the variant_name_index to find the corresponding variant.
     //
     // There's no advantage in our implementation between knowing values
     // statically vs dynamically, so we check the dynamic case first since
--- a/src/adapter/optimizations/variant_lookup.rs
+++ b/src/adapter/optimizations/variant_lookup.rs
@@ -1,0 +1,196 @@
+use std::rc::Rc;
+
+use rustdoc_types::{Id, Item, ItemEnum, VariantKind};
+use trustfall::FieldValue;
+use trustfall::provider::{
+    AsVertex, CandidateValue, ContextIterator, ContextOutcomeIterator, ResolveEdgeInfo, VertexInfo,
+    VertexIterator, resolve_neighbors_with,
+};
+
+use crate::hashtables::HashMap;
+
+use super::super::{
+    RustdocAdapter, enum_variant::LazyDiscriminants, origin::Origin, vertex::Vertex,
+};
+
+fn compute_variant_discriminants<'a>(
+    outer_item: &'a Item,
+    item_index: &'a HashMap<Id, Item>,
+) -> Option<Rc<LazyDiscriminants<'a>>> {
+    // Discriminants are only well-defined if either:
+    // - the enum has a defined `repr` binary representation, or
+    // - none of the enum variants contain any fields of their own.
+    let has_repr = outer_item.attrs.iter().any(|attr| match attr {
+        rustdoc_types::Attribute::Repr(attr_repr) => attr_repr.int.is_some(),
+        _ => false,
+    });
+
+    let enum_item = match &outer_item.inner {
+        ItemEnum::Enum(x) => x,
+        _ => unreachable!("Item {outer_item:?} is not an Enum"),
+    };
+
+    let mut has_fields_in_variants = false;
+    let variants = enum_item
+        .variants
+        .iter()
+        .map(|field_id| {
+            let inner = &item_index.get(field_id).expect("missing item").inner;
+            match inner {
+                ItemEnum::Variant(v) => {
+                    match &v.kind {
+                        VariantKind::Plain => {}
+                        VariantKind::Tuple(t) => has_fields_in_variants |= !t.is_empty(),
+                        VariantKind::Struct { fields, .. } => {
+                            has_fields_in_variants |= fields.is_empty()
+                        }
+                    }
+                    v
+                }
+                _ => unreachable!("Item {inner:?} not a Variant"),
+            }
+        })
+        .collect();
+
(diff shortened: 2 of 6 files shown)


src/adapter/edges.rs
```diff
@@ -527,84 +531,244 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
...
-                Origin::CurrentCrate => &current_crate.own_crate.inner.index,
-                Origin::PreviousCrate => {
-                    &previous_crate
-                        .expect("no previous crate provided")
-                        .own_crate
-                        .inner
-                        .index
-                }
-            };
+        "variant" => {
```
obi1kenobi, line 453: Please follow the current pattern of putting optimization-capable edge resolver implementations in a new file in the `optimizations` submodule, so they don't make `edges.rs` any more crowded than it already is.

src/adapter/edges.rs
```diff
@@ -527,84 +531,244 @@ pub(super) fn resolve_variant_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
     }
 }
 
+fn compute_variant_discriminants<'a>(
+    outer_item: &'a Item,
+    item_index: &'a HashMap<Id, Item>,
+) -> Option<Rc<LazyDiscriminants<'a>>> {
```
obi1kenobi: Are we duplicating the discriminants logic here?

This is at quite some risk of drifting in a way that would be really painful to discover: if the "regular" discriminants computation changes but this isn't updated, then our returned discriminants would only be incorrect when the index is applied to queries.

I've had to debug a couple of those bugs already, and it was extremely painful. I'd really prefer if we can make it impossible for us to have another one.

src/indexed_crate.rs
```diff
@@ -211,6 +211,9 @@ pub struct IndexedCrate<'a> {
     /// index: kind of public top-level item -> `IndexMap<Id, &'a Item>` of that kind
     pub(crate) pub_item_kind_index: PubItemKindIndex<'a>,
 
+    /// index: enum id + variant name -> Variant
+    pub(crate) variant_name_index: Option<HashMap<(Id, &'a str), (&'a Item, usize)>>,
```
obi1kenobi: Make sure to document the value precisely. Presumably the `usize` is the index of the variant inside its enum definition?

src/indexed_crate.rs
```diff
@@ -602,6 +606,10 @@ impl<'a> IndexedCrate<'a> {
         value.fn_owner_index = Some(fn_owner_index);
         value.export_name_index = Some(build_export_name_index(&crate_.index));
 
+        // TODO: Could we use pub_item_kind_index to construct this index?
+        // I expect it would be faster, but would it lose information?
+        value.variant_name_index = Some(build_variant_name_index(&crate_.index));
```
obi1kenobi: It would lose information because it would only include `pub enum` items instead of all `enum` items.

That _might_ be acceptable? It's a question of:
- Is the construction time meaningfully different?
- Are there queries we'd want to support that would be unable to use the narrower index?
- Would it be worth the increase in code complexity required to only apply the index when we know the enum must be `pub`, either directly via a filter on `visibility_limit` that requires it to be `pub`, or via a mandatory `importable_path` edge? Or if we only apply the filter in one of those two cases, if the missed optimization is worth missing to speed up the index construction?

My gut feeling is "not worth it" but would be open to be convinced otherwise.

src/indexed_crate.rs
```diff
@@ -765,6 +773,39 @@ fn build_export_name_index(index: &HashMap<Id, Item>) -> HashMap<&str, &Item> {
...
+        _ => None,
+    })
+    .flat_map(|(id, enum_item)| {
+        #[cfg(feature = "rayon")]
+        let base_iter = enum_item.variants.par_iter();
+        #[cfg(not(feature = "rayon"))]
+        let base_iter = enum_item.variants.iter();
+
+        base_iter.copied().enumerate().map(move |(i, var_id)| {
+            let var = index.get(&var_id).expect("Variant id should be correct.");
```
obi1kenobi: rustdoc JSON occasionally has bugs that result in dangling IDs, and some of those issues are still open. We _shouldn't_ hit this panic, but we _might_.

Slight preference for using `filter_map()` and omitting invalid items from the index instead of exploding, just so we don't cause ourselves extra support burden if such a bug affects our users.

src/indexed_crate.rs
```diff
@@ -765,6 +773,39 @@ fn build_export_name_index(index: &HashMap<Id, Item>) -> HashMap<&str, &Item> {
...
+            (
+                (
+                    id,
+                    var.name
+                        .as_ref()
+                        .expect("Variant should have a name.")
+                        .as_str(),
+                ),
+                (var, i),
+            )
```
obi1kenobi: This is very difficult to read.

Please consider giving each intermediate result an intuitive name, and then just using the return expression to construct the tuples: `((enum_id, variant_name), (variant_id, index))` or similar.

In general, I'm not a fan of using super short variable names like `var`. They save a split second over using something more descriptive like `variant_item`, but the code is then much more difficult to re-read and re-understand 6 months or 2 years later. So whenever possible, please make the code not just correct on day 1 but easy to understand and extend 2 years later!

src/adapter/edges.rs
CLIDragon: As far as I can tell, this is the only instance of the discriminants logic. I have moved it to the new optimizations file, since it is the only place it is currently used, but I'm not sure it's the best place for it, code-grouping wise.

src/indexed_crate.rs
CLIDragon: I have made the change, but I'm not convinced that it's a good idea. If dangling ids do exist, then it's likely that they will cause a false positive in `enum_variant_missing`. Then, instead of being able to point to the backtrace and easily identify the cause as dangling ids in rustdoc, there will be no clear signal that determines the cause.

src/adapter/edges.rs
obi1kenobi: Ah I see, I missed that it got moved instead of duplicated. I'm also not sure if that's the best place, but if it's the only place where it's used then maybe it's okay for now.

obi1kenobi:
Broadly looks in decent shape, just needs some more polishing to improve maintainability: using `unreachable!()` with a good comment that includes the problematic values instead of a bare `panic!()`, using `.expect()` (or even `.unwrap_or_else(|| unreachable!("format string: {data:?}"))`) instead of `.unwrap()`, etc.

src/indexed_crate.rs
CLIDragon: Construction time doesn't appear to be significantly different, so it doesn't appear to be worth making the change.

CLIDragon:
This is ready for review. I'm not happy with the code duplication in `build_variant_name_index` but I can't see a way around it.

CLIDragon:
After some more analysis, this isn't producing the performance results that I had initially hoped for. I would like to see whether there are other ways of speeding up the same queries that are more impactful.

**EDIT:** After a little more research, the only really discontinuous query is `enum_unit_variant_changed_kind`.

CLIDragon:
After some review, resolving variants does appear to have the intended effect - it's simply smaller in magnitude than I was hoping for.

CLIDragon:
Apologies for the churn. While the variant resolution itself appears to be correct, I'm currently running into an issue where tracing reports that this should be reducing `enum_unit_variant_changed_kind` runtime by nearly 90%, when in practice it only reduces it by ~10%.

obi1kenobi:
No worries at all. If you feel a second pair of eyes on the trace might be helpful, I'm happy to take a look too!

CLIDragon:
Running `c-s-c` again, I find that this does produce the correct speedup, so I'm guessing that I mixed up different experiments when I initially recorded the changes. 

@obi1kenobi Could you confirm that, when running on 4 threads in release mode on `aws_sdk_ec2`, `enum_unit_variant_changed_kind` initially takes around `0.38` seconds and after including this patch, takes around `0.02` seconds?

obi1kenobi:
Yes, I'm seeing a 10x speedup on that lint with this branch patched in than without it.

```
unpatched:
        PASS [   0.885s]       major        enum_unit_variant_changed_kind
        PASS [   0.858s]       major        enum_unit_variant_changed_kind        
        PASS [   0.848s]       major        enum_unit_variant_changed_kind

patched w/ this branch:
        PASS [   0.082s]       major        enum_unit_variant_changed_kind
        PASS [   0.074s]       major        enum_unit_variant_changed_kind
        PASS [   0.083s]       major        enum_unit_variant_changed_kind
```

CLIDragon:
Awesome, this should (finally) be ready to merge then.

src/adapter/optimizations/variant_lookup.rs
```diff
@@ -0,0 +1,196 @@
...
+        .collect();
+
+    if has_repr || !has_fields_in_variants {
+        Some(Rc::new(LazyDiscriminants::new(variants)))
+    } else {
+        None
+    }
+}
+
+pub(crate) fn resolve_enum_variants_slow_path<'a>(
```
obi1kenobi: Do we need all these helpers to be `pub(crate)` or can we just make them local to this file?

src/adapter/optimizations/variant_lookup.rs
```diff
@@ -0,0 +1,196 @@
...
+            resolve_enum_variant_by_name(origin, item_vertex, name, variant_name_index, item_index)
+        }
+        CandidateValue::Multiple(values) => Box::new(values.into_iter().flat_map(move |name| {
+            resolve_enum_variant_by_name(origin, item_vertex, name, variant_name_index, item_index)
+        })),
+        _ => resolve_enum_variants_slow_path(origin, item_vertex, item_index),
+    }
+}
+
+pub(crate) fn resolve_enum_variant<'a, V: AsVertex<Vertex<'a>> + 'a>(
```
obi1kenobi, line 130: AFAICT this is the only function that truly needs `pub(crate)` here, right?

obi1kenobi:
Aside from the `pub(crate)` functions, my remaining concern is that right now I don't think we have any test cases that exercise the statically-known path (`@filter` with a `$variable`), and IIRC we don't even have any code that exercises the dynamically-known path (`@filter` with a `%tag`) either.

So while I believe the implementation is correct, it's concerning that it might not be (or might break in the future) and we'd never know. Please consider adding test cases that hit both code paths, and produce a non-empty set of results.

To make the `@filter` with a `%tag` case, you could match a variant to itself in an optimized fashion like so:
```graphql
{
    Crate {
        item {
            ... on Enum {
                variant {
                    name @tag @output
                }
                variant {
                    other: name @filter(op: "=", value: ["%name"]) @output
                }
            }
        }
    }
}
```
This is a bit of a silly query, but it should still be optimized and not be O(n^2).

You could add a new test crate for the optimizations specifically, or extend one of the existing test crates that deals with enum variants — your call.

obi1kenobi:
Heads up: since Rust 1.89 was released ~yesterday, I'm hoping to release a new cargo-semver-checks very soon, ideally this weekend time permitting.

It's not a big deal if we can't get this PR in in time, and I don't want you to re-plan your entire weekend over it. There will be more releases! I just wanted to mention it so you aren't surprised.

CLIDragon:
I believe the improved queries all trigger the dynamic path, e.g.`enum_variant_missing` 

```graphql
CrateDiff {
    baseline {
        item {
            ... on Enum {
                visibility_limit @filter(op: "=", value: ["$public"]) @output

                importable_path {
                    path @output @tag
                    public_api @filter(op: "=", value: ["$true"])
                }

                variant {
                    kind: __typename @output

                    variant_name: name @output @tag
                    public_api_eligible @filter(op: "=", value: ["$true"])

                    span_: span @optional {
                        filename @output
                        begin_line @output
                        end_line @output
                    }
                }
            }
        }
    }
    current {
        item {
            ... on Enum {
                visibility_limit @filter(op: "=", value: ["$public"])
                enum_name: name @output

                importable_path {
                    path @filter(op: "=", value: ["%path"])
                    public_api @filter(op: "=", value: ["$true"])
                }

                variant @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
                    name @filter(op: "=", value: ["%variant_name"])
                }
            }
        }
    }
}
```

CLIDragon:
I've added tests and removed the excess `pub(crate)`s.

src/adapter/optimizations/variant_lookup.rs
```diff
@@ -0,0 +1,196 @@
...
+    adapter: &'a RustdocAdapter<'a>,
+    contexts: ContextIterator<'a, V>,
+    resolve_info: &ResolveEdgeInfo,
+) -> ContextOutcomeIterator<'a, V, VertexIterator<'a, Vertex<'a>>> {
+    // If the name of the variant is required to be a particular value, we can
+    // use the variant_name_index find the corresponding variant.
```
obi1kenobi:
```suggestion
    // use the variant_name_index to find the corresponding variant.
```

src/indexed_crate.rs
```diff
@@ -211,6 +211,10 @@ pub struct IndexedCrate<'a> {
     /// index: kind of public top-level item -> `IndexMap<Id, &'a Item>` of that kind
     pub(crate) pub_item_kind_index: PubItemKindIndex<'a>,
 
+    /// index: enum id + variant name -> variant, index of variant inside enum.
+    #[allow(clippy::type_complexity)]
```
obi1kenobi: Let's use
```suggestion
    #[expect(clippy::type_complexity)]
```
so that if this ever stops triggering `type_complexity`, we are reminded to remove the suppression.

obi1kenobi (approved):
Awesome work!

PR #927, opened 2025-07-28
Introduce new function, `crate_at_origin` to simplify common code path.

CLIDragon:
**Purpose of Change**
This is something that is done a lot and is basically copy-pasted between call sites. Especially in lookup code, such as in #926, the match can harm readability.

**Describe the Change**
Introduce a new function `crate_at_origin`. Since it's `pub(crate)`, it won't affect the API, but it can make internal code easier to read. I have changed some of the uses to demonstrate the changes.

**Describe Alternatives**
Perhaps `package_index_at_origin` may be a better name.

### DIFF
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -53,10 +53,16 @@ impl<'a> RustdocAdapter<'a> {
         &SCHEMA
     }
 
-    pub(crate) fn crate_at_origin(&self, origin: &Origin) -> Option<&'a PackageIndex<'a>> {
+    /// Returns the crate at the given origin.
+    ///
+    /// Panics if `Origin::PreviousCrate` is passed and there is no previous crate.
+    #[inline]
+    pub(crate) fn crate_at_origin(&self, origin: Origin) -> &'a PackageIndex<'a> {
         match origin {
-            Origin::CurrentCrate => Some(self.current_crate),
-            Origin::PreviousCrate => self.previous_crate,
+            Origin::CurrentCrate => self.current_crate,
+            Origin::PreviousCrate => self
+                .previous_crate
+                .expect("previous crate was not provided"),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/adapter/mod.rs
```diff
@@ -52,6 +52,13 @@ impl<'a> RustdocAdapter<'a> {
     pub fn schema() -> &'static Schema {
         &SCHEMA
     }
+
+    pub(crate) fn crate_at_origin(&self, origin: &Origin) -> Option<&'a PackageIndex<'a>> {
```
CLIDragon: I would prefer to move the `expect` inside and just return `&'a PackageIndex<'a>`, but I'm not sure how to communicate clearly to callers that the function will panic if requesting a crate that doesn't exist. OTOH, every current `match origin` expects `previous_crate` to exist anyway, so it might be okay to just leave a doc comment.

obi1kenobi: I agree, let's move the `expect()` inside and mention in the doc comment that the function will panic if the baseline crate is required but not set. That should be a bug anyway so it won't be that surprising to get a panic, I think.

obi1kenobi: I think `Origin` might be `Copy` so we might not need `&Origin`, `Origin` would do.

obi1kenobi: Probably a good idea to suggest to the compiler that it inline this function.
```suggestion
    #[inline]
    pub(crate) fn crate_at_origin(&self, origin: &Origin) -> Option<&'a PackageIndex<'a>> {
```

obi1kenobi (commented):
This looks pretty good. It used to be the case that the
```rust
let current_crate = adapter.current_crate;
let previous_crate = adapter.previous_crate;
```
bits were required when the functions only required `&RustdocAdapter<'a>` and not `&'a RustdocAdapter<'a>`. But since we changed that, this more elegant option is available and we should use it.

Feel free to change the remaining call sites if you'd like, either here or in a separate PR — I'd be happy to merge it either way.

obi1kenobi (approved):

obi1kenobi:
Unfortunately latest nightly moved to rustdoc v55. I'll have to make the `rustdoc-v55` branch that accounts for the new attribute representation changes, then re-point this to merge into `rustdoc-v55` so that CI passes.

I'll take care of it when I get home from my trip on Monday (probably Tuesday in your time zone).

PR #928, opened 2025-08-04
Introduce new function, `crate_at_origin` to simplify common code path. (#927)

obi1kenobi:
**Purpose of Change**
This is something that is done a lot and is basically copy-pasted
between call sites. Especially in lookup code, such as in #926, the
match can harm readability.

**Describe the Change**
Introduce a new function `crate_at_origin`. Since it's `pub(crate)`, it
won't affect the API, but it can make internal code easier to read. I
have changed some of the uses to demonstrate the changes.

**Describe Alternatives**
Perhaps `package_index_at_origin` may be a better name.

---------

Co-authored-by: Predrag Gruevski <2348618+obi1kenobi@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        toolchain: ["nightly"]
+        toolchain: ["beta"]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


obi1kenobi:
I'm going to keep an eye on whether rustdoc v54 or v55 ends up in the new Rust beta this week, and then either merge this (if v54) or abandon it and deprecate v54 (if beta has v55 instead).

obi1kenobi:
Alright, Rust 1.90 beta is on rustdoc-v55, so the v54 lineage is officially no longer supported.

PR #935, opened 2025-08-09
Implement a tracing adapter

CLIDragon:
**Describe the Change**
Add a new struct, `TracingRustdocAdapter` that traces queries at the level of `trustfall`. The design is borrowed from `trustfall`'s internal tracing used for tests.

**Purpose of Change**
`trustfall` by default provides limited mechanisms to examine it's execution, which makes performance analysis difficult. 

**Open Questions**
- How should the adapter be exposed to users?
- What is the best way to test that traces remain valid?
- Is the tracer as efficient as possible, or can performance improvements be made (within reason)?
- Is there missing information?

### DIFF
--- a/src/adapter/tracer/mod.rs
+++ b/src/adapter/tracer/mod.rs
@@ -1,4 +1,4 @@
-pub mod ptrace;
+mod ptrace;
 
 #[cfg(test)]
 mod tests;
--- a/src/adapter/tracer/mod.rs
+++ b/src/adapter/tracer/mod.rs
@@ -1,0 +1,4 @@
+mod ptrace;
+
+#[cfg(test)]
+mod tests;
--- a/src/adapter/tracer/mod.rs
+++ b/src/adapter/tracer/mod.rs
@@ -1,4 +1,4 @@
-mod ptrace;
+pub mod ptrace;
 
 #[cfg(test)]
 mod tests;
--- a/src/adapter/tracer/ptrace.rs
+++ b/src/adapter/tracer/ptrace.rs
@@ -18,53 +18,59 @@ use trustfall::{
 /// values.
 #[derive(Debug, Clone)]
 pub struct ExpHistogram {
-    buckets: Vec<u32>,
-    count: u32,
+    buckets: [u32; 16],
 }
+
+pub const BOUNDARIES: [u64; 16] = [
+    100,
+    300,
+    1000,
+    3000,
+    10000,
+    30000,
+    100000,
+    300000,
+    1000000,
+    3000000,
+    10000000,
+    30000000,
+    100000000,
+    300000000,
+    1000000000,
+    u64::MAX,
+];
 
 impl ExpHistogram {
     /// Create a new histogram.
     pub fn new() -> ExpHistogram {
-        ExpHistogram {
-            buckets: vec![0; 16],
-            count: 0,
-        }
+        ExpHistogram { buckets: [0; 16] }
     }
 
     /// Add a value to the histogram.
     pub fn add(&mut self, num: u64) {
-        self.count += 1;
-
-        if num <= 100 {
-            self.buckets[0] += 1;
-            return;
+        for (i, lim) in BOUNDARIES.iter().enumerate() {
+            if num <= *lim {
+                self.buckets[i] = self.buckets[i].saturating_add(1);
+                break;
+            }
         }
-
-        let mut log = (num as f64).log10();
-        if log == log.floor() {
-            // The bucket is 1 off for exact powers of 10 without correction.
-            log -= 0.4;
-        }
-        let bucket = ((log * 2.0 - 3.0).floor() as usize).min(15);
-
-        self.buckets[bucket] += 1;
     }
 
     /// Returns the largest value that will be accepted into each bucket.
-    pub fn boundaries(&self) -> Vec<u64> {
-        (0..=14)
-            .map(|x| 10.0_f64.powf(((x as f64) / 2.0) + 2.0) as u64)
-            .chain(std::iter::once(u64::MAX))
-            .collect()
+    pub fn boundaries(&self) -> &'static [u64; 16] {
+        &BOUNDARIES
     }
 
     /// Returns the number of values stored in the histogram.
     pub fn count(&self) -> u32 {
-        self.count
+        self.buckets()
+            .iter()
+            .copied()
+            .fold(0, |acc, num| acc.saturating_add(num))
     }
 
     /// Returns the count of each bucket
-    pub fn buckets(&self) -> &Vec<u32> {
+    pub fn buckets(&self) -> &[u32; 16] {
         &self.buckets
     }
 }
@@ -104,7 +110,7 @@ impl Summary {
 
         self.min = self.min.min(duration);
         self.max = self.max.max(duration);
-        self.sum += duration;
+        self.sum = self.sum.saturating_add(duration);
     }
 
     /// Returns the number of items that have been processed.
@@ -134,6 +140,7 @@ impl Summary {
 
     /// Returns the mean duration.
     pub fn mean(&self) -> Duration {
+        // This cannot panic since count() is always at least 1.
         self.sum / self.count()
     }
 }
@@ -189,7 +196,6 @@ impl Default for Tracer {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
-#[allow(clippy::enum_variant_names)] // Names match the adapter functions 
 pub enum FunctionCall {
     ResolveProperty(Vid, Arc<str>, Arc<str>), // vertex ID + type name + name of the property
     ResolveNeighbors(Vid, Arc<str>, Eid),     // vertex ID + type name + edge ID
@@ -236,8 +242,9 @@ where
     PerfSpanIter { inner, post_action }
 }
 
-/// Traces an inner adapter. Constructed with new(). Each tracer is only valid
-/// for one query. When switching between queries, ensure you first call finish(),
+/// Traces an inner adapter. Constructed with [TracingAdapter::new()]. Each
+/// tracer is only valid for one query. When switching between queries, if you
+/// use the same adapter, ensure you first call [TracingAdapter::finish()],
 /// otherwise the new operations will be traced with the old.
 #[derive(Debug, Clone)]
 pub struct TracingAdapter<'vertex, AdapterT>
@@ -299,14 +306,22 @@ where
         property_name: &Arc<str>,
         resolve_info: &ResolveInfo,
     ) -> ContextOutcomeIterator<'vertex, V, FieldValue> {
+        // TODO: Move this comment to a better location, to make it clear it applies
+        //      to all resolve_* functions (except resolve_starting_vertices).
         // For each resolution we want to know:
         // 1. What are we resolving?
         // 2. How long did the resolution take?
-        // 3. How many times did we resolve the same property?
+        // 3. How many times was a particular (vid, type, property) triple resolved?
         //
         // We are not collecting:
-        // 1. Number of times a function is called.
-        // 2. Whether or not any results were returned.
+        // (1) Number of times a function is called, and time spent constructing
+        //     the iterators.
+        //
+        //     The number of times any given resolution function is called is
+        //     an implementation detail of trustfall, and since most time is spent
+        //     resolving iterators, tracking it isn't valuable.
+        //
+        // (2) Whether or not the returned iterator is empty.
 
         let call_id = FunctionCall::ResolveProperty(
             resolve_info.vid(),
@@ -351,7 +366,7 @@ where
         resolve_info: &ResolveEdgeInfo,
     ) -> ContextOutcomeIterator<'vertex, V, VertexIterator<'vertex, Self::Vertex>> {
         // Inner and outer call times are often quite different, so they need to
-        // be stored separately.
+        // be tracked separately.
         let call_id = FunctionCall::ResolveNeighbors(
             resolve_info.origin_vid(),
             type_name.clone(),
--- a/src/adapter/tracer/ptrace.rs
+++ b/src/adapter/tracer/ptrace.rs
@@ -15,7 +15,7 @@ use trustfall::{
 
 /// A simple histogram that stores 15 roughly exponentially increasing buckets of
 /// values, from 0 to 1 billion, followed by a final bucket to store numbers greater
-/// 1 billion.
+/// than 1 billion.
 #[derive(Clone)]
 pub struct ExpHistogram {
     buckets: [u32; 16],
@@ -254,9 +254,9 @@ where
     PerfSpanIter { inner, post_action }
 }
 
-/// Traces an inner adapter. Constructed with [TracingAdapter::new()]. Each
+/// Traces an inner adapter. Constructed with [`TracingAdapter::new()`]. Each
 /// tracer is only valid for one query. When switching between queries, if you
-/// use the same adapter, ensure you first call [TracingAdapter::finish()],
+/// use the same adapter, ensure you first call [`TracingAdapter::finish()`],
 /// otherwise the new operations will be traced with the old.
 #[derive(Debug, Clone)]
 pub struct TracingAdapter<'vertex, AdapterT>
--- a/src/adapter/tracer/ptrace.rs
+++ b/src/adapter/tracer/ptrace.rs
@@ -3,8 +3,6 @@ use std::{
     time::Duration,
 };
 
-use kll_rs::KllDoubleSketch;
-
 use std::time::Instant;
 
 use trustfall::{
@@ -15,9 +13,72 @@ use trustfall::{
     },
 };
 
-#[derive(Debug)]
+/// A simple histogram that stores 15 roughly exponentially increasing buckets of
+/// values, from 0 to 1 billion, followed by a final bucket to store greater
+/// values.
+#[derive(Debug, Clone)]
+pub struct ExpHistogram {
+    buckets: Vec<u32>,
+    count: u32,
+}
+
+impl ExpHistogram {
+    /// Create a new histogram.
+    pub fn new() -> ExpHistogram {
+        ExpHistogram {
+            buckets: vec![0; 16],
+            count: 0,
+        }
+    }
+
+    /// Add a value to the histogram.
+    pub fn add(&mut self, num: u64) {
+        self.count += 1;
+
+        if num <= 100 {
+            self.buckets[0] += 1;
+            return;
+        }
+
+        let mut log = (num as f64).log10();
+        if log == log.floor() {
+            // The bucket is 1 off for exact powers of 10 without correction.
+            log -= 0.4;
+        }
+        let bucket = ((log * 2.0 - 3.0).floor() as usize).min(15);
+
+        self.buckets[bucket] += 1;
+    }
+
+    /// Returns the largest value that will be accepted into each bucket.
+    pub fn boundaries(&self) -> Vec<u64> {
+        (0..=14)
+            .map(|x| 10.0_f64.powf(((x as f64) / 2.0) + 2.0) as u64)
+            .chain(std::iter::once(u64::MAX))
+            .collect()
+    }
+
+    /// Returns the number of values stored in the histogram.
+    pub fn count(&self) -> u32 {
+        self.count
+    }
+
+    /// Returns the count of each bucket
+    pub fn buckets(&self) -> &Vec<u32> {
+        &self.buckets
+    }
+}
+
+impl Default for ExpHistogram {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// A summary of key timing statistics.
+#[derive(Debug, Clone)]
 pub struct Summary {
-    sketch: KllDoubleSketch,
+    hist: ExpHistogram,
     min: Duration,
     max: Duration,
     sum: Duration,
@@ -26,13 +87,11 @@ pub struct Summary {
 impl Summary {
     // By initialising with a duration, we don't require min/max to be options.
     pub fn new(duration: Duration) -> Summary {
-        // Based on https://datasketches.apache.org/docs/KLL/KLLAccuracyAndSize.html
-        // a K of 200 should give ~1.33% error.
-        let mut sketch = KllDoubleSketch::new_with_k(200).unwrap();
-        sketch.update(duration.as_nanos() as f64);
+        let mut hist = ExpHistogram::new();
+        hist.add(duration.as_nanos() as u64);
 
         Summary {
-            sketch,
+            hist,
             min: duration,
             max: duration,
             sum: duration,
@@ -41,7 +100,7 @@ impl Summary {
 
     /// Add a new time to the summary.
     pub fn update(&mut self, duration: Duration) {
-        self.sketch.update(duration.as_nanos() as f64);
+        self.hist.add(duration.as_nanos() as u64);
 
         self.min = self.min.min(duration);
         self.max = self.max.max(duration);
@@ -49,8 +108,8 @@ impl Summary {
     }
 
     /// Returns the number of items that have been processed.
-    pub fn count(&self) -> u64 {
-        self.sketch.get_n()
+    pub fn count(&self) -> u32 {
+        self.hist.count()
     }
 
     /// Returns the total time
@@ -68,10 +127,14 @@ impl Summary {
         self.max
     }
 
-    /// Returns the quantile (0 < quant < 1) time.
-    pub fn quantile(&self, quant: f64) -> f64 {
-        assert!(0.0 < quant && quant < 1.0);
-        self.sketch.get_quantile(quant)
+    /// Returns the exponential histogram with times in nanoseconds.
+    pub fn histogram(&self) -> &ExpHistogram {
+        &self.hist
+    }
+
+    /// Returns the mean duration.
+    pub fn mean(&self) -> Duration {
+        self.sum / self.count()
     }
 }
 
@@ -80,9 +143,9 @@ impl Summary {
 /// This struct is intended for use inside of a TracingAdapter.
 /// Operations must be recorded sequentially in chronological order.
 /// Recording out-of-order operations will lead to invalid state.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Tracer {
-    calls: BTreeMap<FunctionCall, Summary>,
+    pub calls: BTreeMap<FunctionCall, Summary>,
 
     // When we measure the time of an iterator, we also measure the time spent
     // evaluating its inputs. We must therefore subtract this time out when
@@ -116,6 +179,12 @@ impl Tracer {
     /// Get the duration of the last input. Panics if the duration is None.
     pub fn get_last_input_duration(&self) -> Duration {
         self.last_input_duration.unwrap()
+    }
+}
+
+impl Default for Tracer {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
@@ -186,15 +255,16 @@ where
     AdapterT: Adapter<'vertex>,
     AdapterT::Vertex: Debug + Clone + 'vertex,
 {
-    pub fn new(adapter: AdapterT, tracer: Rc<RefCell<Tracer>>) -> Self {
+    pub fn new(adapter: AdapterT) -> Self {
         Self {
-            tracer,
+            tracer: Rc::new(RefCell::new(Tracer::new())),
             inner: adapter,
             _phantom: PhantomData,
         }
     }
 
-    pub fn finish(self) -> Tracer {
+    /// Finalise the trace and return it.
+    pub fn finish(&self) -> Tracer {
         // Ensure nothing is reading the trace i.e. we can safely stop interpreting.
         let trace_ref = self.tracer.borrow_mut();
         let new_trace = Tracer::new();
--- a/src/adapter/tracer/ptrace.rs
+++ b/src/adapter/tracer/ptrace.rs
@@ -1,10 +1,9 @@
 use std::{
-    cell::RefCell, collections::BTreeMap, fmt::Debug, marker::PhantomData, num::NonZeroU32, rc::Rc,
-    sync::Arc, time::Duration,
+    cell::RefCell, collections::BTreeMap, fmt::Debug, marker::PhantomData, rc::Rc, sync::Arc,
+    time::Duration,
 };
 
 use kll_rs::KllDoubleSketch;
-use serde::{Deserialize, Serialize};
 
 use std::time::Instant;
 
@@ -33,7 +32,7 @@ impl Summary {
         sketch.update(duration.as_nanos() as f64);
 
         Summary {
-            sketch: sketch,
+            sketch,
             min: duration,
             max: duration,
             sum: duration,
@@ -102,7 +101,7 @@ impl Tracer {
 
     /// Record an operation.
     pub fn record_time(&mut self, call_id: &FunctionCall, duration: Duration) {
-        if let Some(summary) = self.calls.get_mut(&call_id) {
+        if let Some(summary) = self.calls.get_mut(call_id) {
             summary.update(duration);
         } else {
             self.calls.insert(call_id.clone(), Summary::new(duration));
@@ -121,6 +120,7 @@ impl Tracer {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[allow(clippy::enum_variant_names)] // Names match the adapter functions 
 pub enum FunctionCall {
     ResolveProperty(Vid, Arc<str>, Arc<str>), // vertex ID + type name + name of the property
     ResolveNeighbors(Vid, Arc<str>, Eid),     // vertex ID + type name + edge ID
--- a/src/adapter/tracer/ptrace.rs
+++ b/src/adapter/tracer/ptrace.rs
@@ -14,14 +14,15 @@ use trustfall::{
 };
 
 /// A simple histogram that stores 15 roughly exponentially increasing buckets of
-/// values, from 0 to 1 billion, followed by a final bucket to store greater
-/// values.
-#[derive(Debug, Clone)]
+/// values, from 0 to 1 billion, followed by a final bucket to store numbers greater
+/// 1 billion.
+#[derive(Clone)]
 pub struct ExpHistogram {
     buckets: [u32; 16],
 }
 
-pub const BOUNDARIES: [u64; 16] = [
+/// The largest value that will be accepted into each bucket of the histogram.
+pub const HIST_BOUNDARIES: [u64; 16] = [
     100,
     300,
     1000,
@@ -48,7 +49,7 @@ impl ExpHistogram {
 
     /// Add a value to the histogram.
     pub fn add(&mut self, num: u64) {
-        for (i, lim) in BOUNDARIES.iter().enumerate() {
+        for (i, lim) in HIST_BOUNDARIES.iter().enumerate() {
             if num <= *lim {
                 self.buckets[i] = self.buckets[i].saturating_add(1);
                 break;
@@ -58,7 +59,7 @@ impl ExpHistogram {
 
     /// Returns the largest value that will be accepted into each bucket.
     pub fn boundaries(&self) -> &'static [u64; 16] {
-        &BOUNDARIES
+        &HIST_BOUNDARIES
     }
 
     /// Returns the number of values stored in the histogram.
@@ -72,6 +73,17 @@ impl ExpHistogram {
     /// Returns the count of each bucket
     pub fn buckets(&self) -> &[u32; 16] {
         &self.buckets
+    }
+}
+
+impl Debug for ExpHistogram {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "ExpHistogram {{ buckets: {:?}, count: {:?} }}",
+            &self.buckets,
+            self.count()
+        )
     }
 }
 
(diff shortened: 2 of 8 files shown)


obi1kenobi:
I think we should consider two paths here: either aim for something radically more limited (I'll give an example below) or aim for making Trustfall's own tracing API publicly available in a sensible way that lets us offload a lot of the work here to it instead. 

My concern is that this is quite a bit of complexity going into trustfall-rustdoc-adapter, a project that's already challenging to maintain because 3-6 (rustdoc JSON) versions of it see active development at the same time. 

Aiming for something more limited would look like this: instead of having a _general purpose_ tracing framework, have a single tracing adapter that just records call counts and time spent. It can implement Trustfall's adapter trait directly and more or less just do `Box::new(underlying_call().inspect(|_| self.calls[...] += 1)` for example. The goal there is to get us the specific data we know has been useful so far: queries can be slow is something is called a lot, or if each call takes a long time, and we'd get to see that here. 

The alternative is to first try to make a general framework for building such instrumentation and expose it in Trustfall itself, then use it here. This will likely be really hard and take a long time, because API design off of N=1 use cases is really tricky.

I think the first option can let us build toward the second, so it's not an either-or.

The reason I suggest this is because I don't think performance engineering _so far_ has necessitated full trace recordings at the step-by-step level. We basically were just looking at call counts and time taken, and that was plenty. So let's consider shipping a less complex thing that'll be easier to maintain and will check all the boxes we need?

CLIDragon:
The right complexity level for the tracing is something that I put quite a bit of thought into at the start of the project, and I have perhaps been falling prey to the sunk-cost fallacy by continuing with the current design. The main reason for keeping in all the information about the call stack (encoded in `TraceOpType`), was to hopefully be able to recreate it in a format that could be plugged directly into existing analysis tools. Additionally, I kept the framework used by `trustfall` in the hopes it would be easier to integrate back into. However, the design evolved so that there is enough of a discrepancy that `trustfall`'s tracing framework would need to be completely re-written - and if it's done in such a way that it's useful to other users, then it's likely to be a large undertaking. In short, I agree that there's unnecessary complexity in this implementation.

However, with that said, I'm not sure that it's possible to drastically simplify the implementation while keeping the information that I want. In particular, I would like the call count for each function, its runtime, and how many times each inner iterator yields[^1]. I have pushed a version with a simplified implementation that just collects this information. I haven't updated tests, so just considering the `ptrace.rs` file, do you think that it's roughly the right level of implementation complexity? Or do you think it should be reduced further?

To show why I want this particular information, here are the sample outputs of my analysis script. I have attempted to describe anything I think might be unintuitive for each one, but please let me know if there's something that doesn't look right or if I am assuming something is common knowledge that isn't - I've spent so long staring at these files that I have undoubtedly grown used to their quirks. 
```
// Each function is listed by Opid, Signature [num_calls]
// The list beneath each ResolveNeighbors is (calls, count). So Opid(3) 
// was called once and output 62 items, while Opid(9) was called
// 62 times, 4 of which resulted in 0 outputs, 41 resulted in 1 
// and so on for a total of 77 outputs. 
// <1:1> Means that for each input there was exactly one output.
// Note that if we just track the number of calls then it might look 
// like a function is 1:1 when it is actually a 50/50 split between
// two outputs and zero outputs.
=== constructible_struct_adds_field === 
Opid(1) ResolveStartingVertices(Vid(1)) [1]
Opid(2) ResolveNeighbors(Vid(1), "CrateDiff", Eid(1)) [1] <1:1>
Opid(3) ResolveNeighbors(Vid(2), "Crate", Eid(2)) [1]
- [(62, 1)]
- total: 62
Opid(4) ResolveCoercion(Vid(3), "Item", "Struct") [62]
Opid(5) ResolveProperty(Vid(3), "Struct", "visibility_limit") [62]
Opid(6) ResolveProperty(Vid(3), "Struct", "attrs") [62]
Opid(7) ResolveNeighbors(Vid(3), "Struct", Eid(3)) [62] <1:1>
Opid(8) ResolveProperty(Vid(4), "ImportablePath", "public_api") [62]
Opid(9) ResolveNeighbors(Vid(3), "Struct", Eid(4)) [62]
- [(0, 4), (1, 41), (2, 15), (3, 2)]
- total: 77
Opid(10) ResolveProperty(Vid(5), "StructField", "visibility_limit") [77]
```
```
// A table of the amount of time spent in each function.
// ResolveNeighbors are followed by a RNI row, which stands for 
// ResolveNeighborsInner and means the inner iterator calls. 
=== method_parameter_count_changed === 
Time: 413.8782ms
Totals (RNI/Outer/All): 30049237 nanoseconds/115642301 nanoseconds/145691538 nanoseconds
+---------------------------------------------------+----------+-------+------+--------+-----------+
|                     signature                     |  total   | count | mean | median |  percent  |
+===================================================+==========+=======+======+========+===========+
| ResolveNeighbors(Vid(9), "Crate", Eid(9))         | 14225991 | 17680 | 804  | 799    | 9.76%     |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| RNI                                               | 11127698 | 17680 | 629  | 400    | 7.64%     |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| ResolveNeighbors(Vid(10), "ImplOwner", Eid(10))   | 13402413 | 17680 | 758  | 699    | 9.2%      |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| RNI                                               | 904299   | 18040 | 50   | 0      | 0.621%    |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| ResolveNeighbors(Vid(10), "ImplOwner", Eid(11))   | 10277010 | 17680 | 581  | 599    | 7.05%     |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| RNI                                               | 1800000  | 17680 | 101  | 100    | 1.24%     |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| ResolveProperty(Vid(11), "ImportablePath",        | 6849288  | 18040 | 379  | 399    | 4.7%      |
| "path")                                           |          |       |      |        |           |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| ResolveNeighbors(Vid(10), "ImplOwner", Eid(13))   | 6159781  | 17680 | 348  | 301    | 4.23%     |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| RNI                                               | 608500   | 17680 | 34   | 0      | 0.418%    |
+---------------------------------------------------+----------+-------+------+--------+-----------+
| ResolveProperty(Vid(4), "ImportablePath", "path") | 6096541  | 18040 | 337  | 301    | 4.18%     |
+---------------------------------------------------+----------+-------+------+--------+-----------+
... and so on for the rest of the functions.
```

[^1]: I believe that each `resolve_neighbors` currently only creates one inner iterator, but I don't know whether that's intentional, a coincidence, or if I have just not found the right edge case yet.

obi1kenobi:
> I believe that each `resolve_neighbors` currently only creates one inner iterator, but I don't know whether that's intentional, a coincidence, or if I have just not found the right edge case yet.

Let's consider a concrete query, just to make sure we're on the same page. Say we're listing all enum names and their variants. 

The query has one edge from enum to variant, so `resolve_neighbors` will get called only once for that edge. (Also once to get the enums in the crate etc. Each of those calls has a unique Eid in the query IR, and a unique Opid in your trace.)

The `resolve_neighbors` call for enum variants, per its type signature, takes an iterator as input and returns an iterator of iterators. The input specifies which enums we need variants for, and each of them corresponds to a result in the outer returned iterator. The inner iterator returns the variants for that enum. There's a separate inner iterator inside each outer iterator item.

obi1kenobi:
To think about how to simplify this design, try to shift your perspective from thinking about a trace to thinking about a complete execution of a query.

The _order_ of operations doesn't matter there. A trace is a sequence of `Opid` steps. This is valuable in Trustfall because we need to test changes to the query engine that cause changes in execution order and guarantees about lazy evaluation etc. But in the list of properties to capture you said we need, the specific sequence wasn't one of them — because in cargo-semver-checks it doesn't matter, since we run all queries to completion anyway. 

Try to make a recording of the execution that has the query IR at its center, not a trace of execution steps. Think about how you'd overlay the recorded data onto the query syntax visually — you can't do that with a trace, but you can if you record "this is how often this property / edge is resolved" etc.

Don't be afraid to lose some information. For example, we don't need to know exactly how many neighboring edges each specific vertex had — knowing that it's 1 on average and a max of 2 is fine. We don't need to distinguish between 2, 0, 2, 0 and 1, 1, 0, 2 — we wouldn't optimize them any differently. 

The goal here should be "what's the starting set that will be useful, easy to merge, and easy to maintain and extend."

CLIDragon:
> > I believe that each `resolve_neighbors` currently only creates one inner iterator, but I don't know whether that's intentional, a coincidence, or if I have just not found the right edge case yet.
> 
> Let's consider a concrete query, just to make sure we're on the same page. Say we're listing all enum names and their variants.

In this case, the inner iterator I was referring to would be the iterator that actually yields the variants. So for each outer iterator item (corresponding to each enum), there is one inner iterator (corresponding to the variants). I was commenting that it doesn't appear that there are any calls where multiple inner iterators are yielded per outer iterator. In retrospect, I have no way of knowing whether there are or not - the information I collected was not fine-grained enough to be able to determine that.

CLIDragon:
I think I have reduced the collected information to close to the minimum. There's a lot of implementation work left to do, but does the overall structure look like what you were envisioning?

obi1kenobi:
> In this case, the inner iterator I was referring to would be the iterator that actually yields the variants. So for each outer iterator item (corresponding to each enum), there is one inner iterator (corresponding to the variants). I was commenting that it doesn't appear that there are any calls where multiple inner iterators are yielded per outer iterator. In retrospect, I have no way of knowing whether there are or not - the information I collected was not fine-grained enough to be able to determine that.

In this block, you mention:

> where multiple inner iterators are yielded per outer iterator

Per outer _iterator_ not iterator _item_, this is not true in general. If there are multiple enums each with multiple iterators, you'll see multiple outer iterator items (one per enum) and multiple inner iterator items in each (each enum's variants).

But it can be true if we have a single specific enum we're looking for via an index lookup, for example by importable path. Then the optimization logic kicks in and only returns the matching enum, so you'll see the behavior you describe.

There's also of course one inner iterator per outer iterator _item_, per the type signature.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    pub fn calls(&self) -> &BTreeMap<FunctionCall, Vec<Duration>> {
+        &self.calls
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum FunctionCall {
+    ResolveStartingVertices(Vid),              // vertex ID
```
obi1kenobi: This doesn't seem used, since we don't record the first resolution.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,278 @@
...
+                .resolve_property(contexts, type_name, property_name, resolve_info);
+
+        let tracer_ref = self.tracer.clone();
+
+        Box::new(make_iter_with_perf_span(
+            inner_iter,
+            move |(context, value), duration| {
+                tracer_ref
+                    .borrow_mut()
+                    .record_time(call_id.clone(), duration);
```
obi1kenobi: I'm not sure if this `.clone()` (+ similar ones elsewhere) to be able to use the `.entry()` API actually pays off.

On the very first `.record_time()` for a given function, using the `.entry()` API will prevent a double-lookup. Nice!

But the map of entries is quite small, and `.record_time()` is called way more than once. On every subsequent call, it's 1-2 `Arc::clone()` invocations which appear unnecessary because the key will be present.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,278 @@
...
+
+                let tracer_ref_2 = tracer_ref.clone();
+
+                let value = call_id_inner.clone();
+
+                let tapped_neighbor_iter = Box::new(
+                    make_iter_with_perf_span(
+                        neighbor_iter.enumerate(),
```
obi1kenobi: Do we need this `.enumerate()`?

src/adapter/tracer/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod ptrace;
```
obi1kenobi, line 1: I'm not necessarily convinced of the value of the extra `pub` layer of indirection here.

Perhaps `mod ptrace;` plus re-exporting what's intended to be `pub` would suffice?

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,278 @@
...
+};
+
+/// Records and stores operations performed by the adapter.
+///
+/// This struct is intended for use inside of a TracingAdapter.
+/// Operations must be recorded sequentially in chronological order.
+/// Recording out-of-order operations will lead to invalid state.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Tracer {
+    pub calls: BTreeMap<FunctionCall, Vec<Duration>>,
```
obi1kenobi: I'm worried about how big these `Vec` allocations might become. Have you tried recording an `aws-sdk-ec2` execution across all lints and measuring how much memory that takes?

When performance-profiling `cargo-semver-checks`, we should expect to have `O(# of lints)` traces be live concurrently. The number of lints in `cargo-semver-checks` is growing exponentially, so that seems not particularly future-proof.

Do we need _all_ durations, or would some summary statistics suffice? For example, say we had `min, max, sum, count, p50, p75, p90, p95, p99`, where the percentiles use an approximate counting method with reasonable accuracy bounds. Are there ways that lints might be slow that wouldn't be detectable from those statistics, but would be detectable from the full `Vec<Duration>`?

obi1kenobi (commented):
I think this is a big step in the right direction! Some more comments and ideas inline.

CLIDragon:
Storing an approximation of quantiles is surprisingly difficult but [`kll-rs`](https://github.com/homeffjy/kll-rs/tree/master) appears to be the best option. It provides a rusty wrapper over [Apache DataSketches KLL](https://datasketches.apache.org/docs/KLL/KLLAccuracyAndSize.html), which has a nearly minimal size (300 KB for a large query vs  ~30MB storing all values) and microbenchmarks show is only 2-3x slower than storing and sorting a vector.

obi1kenobi:
What crate was being linted in the 30MB recording? And is that per query or total for all lints?

`kll-rs` doesn't look like a good fit. It isn't published on crates.io, it introduces non-Rust code which might have portability problems, and it doesn't seem actively developed so we'd effectively need to become the maintainers.

I think we should instead build a simple exponential histogram and collect counts in each of 16 bins: `<=100ns, <=300us, <=1us, <=3us, <=10us, <=30us, <=100us, <=300us, <=1ms, <=3ms, <=10ms, <=30ms, <=100ms, <=300ms, <=1s, <= +infty`. This uses 128 bytes per histogram and will give us a very solid estimate of the distribution. We don't in practice care to know the difference between our p95 being 2.3ms vs 2.5ms — we'll be looking for bigger fish to fry for a long time to come. We should optimize for the thing we can ship quickly and that we won't regret shipping in the future.

(If we wanted to be perfectionistic, we could use `sqrt(10) ~= 3.16` instead of 3 as the intermediate step between powers of ten.)

CLIDragon:
> What crate was being linted in the 30MB recording? And is that per query or total for all lints?

`aws-sdk-ec2`, for the singular query `method_parameter_count_changed`. The precise figure is probably closer to 26MB (30MB accounting for loss due to excess allocations by the `Vec`), since it takes 2108922 operations, storing 12 bytes each. This could be reduced by storing a `u64` instead of a `Duration` of course, but no matter how compressed it is, 2 million operations will still take some amount of memory to store.

obi1kenobi:
I see. I think storing an exponential histogram + total time in `u64` nanoseconds, for each operation in each query is probably sufficient and preferable.

Each query is give or take 30ish operations we'd care to time. Each operation uses ~136 bytes of metadata, and ~30 * 136 = 4KB. 200 lints * 4KB is a negligible amount of memory, and it will remain negligible even at 2000 lints.

CLIDragon:
I've switch to using an exponential histogram, as suggested. The memory footprint should be about 192 bytes per function call (80 for `Summary`, 48 for `FunctionCall` and another 64 on the heap for `ExpHistogram`).

src/lib.rs
```diff
@@ -18,5 +18,6 @@ pub use cargo_metadata;
 
 pub use {
     adapter::RustdocAdapter,
+    adapter::tracer::ptrace,
```
CLIDragon, line 21: I'm not sure what the best way to export `ptrace` is - any guidance would be appreciated.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+pub struct ExpHistogram {
+    buckets: Vec<u32>,
+    count: u32,
+}
+
+impl ExpHistogram {
+    /// Create a new histogram.
+    pub fn new() -> ExpHistogram {
+        ExpHistogram {
+            buckets: vec![0; 16],
```
obi1kenobi: Since we have a fixed number of buckets, any reason not to just use an array `[u32; 16]` instead of a `Vec`? The pros are that we avoid a layer of indirection and a heap allocation.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+    },
+};
+
+/// A simple histogram that stores 15 roughly exponentially increasing buckets of
+/// values, from 0 to 1 billion, followed by a final bucket to store greater
+/// values.
+#[derive(Debug, Clone)]
+pub struct ExpHistogram {
+    buckets: Vec<u32>,
+    count: u32,
```
obi1kenobi: Isn't the count just the sum of the numbers in the buckets?

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+            self.buckets[0] += 1;
+            return;
+        }
+
+        let mut log = (num as f64).log10();
+        if log == log.floor() {
+            // The bucket is 1 off for exact powers of 10 without correction.
+            log -= 0.4;
+        }
+        let bucket = ((log * 2.0 - 3.0).floor() as usize).min(15);
```
obi1kenobi: All this floating point math is going to be quite slow, and is going to be a bottleneck in itself. Computing even one `log` is going to be massively more expensive than most Trustfall property lookups, for example.

We'd strongly prefer imperfect bucket sizes over floating point math in the hot path. We'd even prefer buckets that are just `1 * 10^N` and `3 * 10^N`, meaning the step size varies between 3 and 10/3 — in real world uses the buckets are often even less evenly spaced than even that.

I'd recommend:
- precompute the bucket cutoffs and store them in a `pub const` of type `[u64; 16]`
- iterate over the cutoffs from smallest to largest to find the right bucket index
- use `.saturating_add()` not `+= 1` because we don't want to panic on overflow

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+        }
+    }
+
+    /// Add a new time to the summary.
+    pub fn update(&mut self, duration: Duration) {
+        self.hist.add(duration.as_nanos() as u64);
+
+        self.min = self.min.min(duration);
+        self.max = self.max.max(duration);
+        self.sum += duration;
```
obi1kenobi: Let's use `saturating_add()` again, to avoid possible panics.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+
+    /// Returns the exponential histogram with times in nanoseconds.
+    pub fn histogram(&self) -> &ExpHistogram {
+        &self.hist
+    }
+
+    /// Returns the mean duration.
+    pub fn mean(&self) -> Duration {
+        self.sum / self.count()
+    }
```
obi1kenobi, line 157: I'd add a note that this can't panic because we're always guaranteed to have at least one sample.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+}
+
+/// A summary of key timing statistics.
+#[derive(Debug, Clone)]
+pub struct Summary {
+    hist: ExpHistogram,
+    min: Duration,
+    max: Duration,
+    sum: Duration,
+}
```
obi1kenobi, line 103: Consider implementing `Display` and possibly hand-implementing `Debug` for both `Summary` and `ExpHistogram`.

As we debug workloads, we'll find ourselves staring at a lot of histograms and summaries. It'd be wonderful to have a high quality, thoughtfully put together representation instead of just `hist: ExpHistogram(buckets=[0, 0, 0, 0, 123, 34568, 3456, 0, 0, 0, 0, 0, 0, 0, 0, 0])` which is _technically accurate_ but I still have to count zeroes to figure out how long things actually took.

Also consider how `ExpHistogram` and `Summary` would get `Debug` or `Display`-printed when wrapped in the data structures of a full trace.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+        &self,
+        contexts: ContextIterator<'vertex, V>,
+        type_name: &Arc<str>,
+        property_name: &Arc<str>,
+        resolve_info: &ResolveInfo,
+    ) -> ContextOutcomeIterator<'vertex, V, FieldValue> {
+        // For each resolution we want to know:
+        // 1. What are we resolving?
+        // 2. How long did the resolution take?
+        // 3. How many times did we resolve the same property?
```
obi1kenobi: Clarification nit to avoid confusion, since in some cases the same property will be resolved more than once in the same query (e.g. `@filter @output` on the same property). That's not what this point implies we're tracking.
```suggestion
        // 3. For how many vertices did we resolve that property?
```

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+        //
+        // We are not collecting:
+        // 1. Number of times a function is called.
+        // 2. Whether or not any results were returned.
```
obi1kenobi: I'm confused by these two. Which function is this referring to?

There's exactly one function call at this layer of abstraction: the one to `resolve_property()`, and it always returns an iterator by the API contract. Whether the wrapped adapter internally implements `resolve_property()` by calling a function for each element of the iterator or not is unknown here, and couldn't possibly be tracked.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+
+fn make_iter_with_perf_span<I, T, F>(inner: I, post_action: F) -> PerfSpanIter<I, T, F>
+where
+    I: Iterator<Item = T>,
+    F: Fn(T, Duration) -> T,
+{
+    PerfSpanIter { inner, post_action }
+}
+
+/// Traces an inner adapter. Constructed with new(). Each tracer is only valid
```
obi1kenobi: When referring to Rust items in doc comments, always use rustdoc references like this:
```suggestion
/// Traces an inner adapter. Constructed with [`TracingAdapter::new()`]. Each tracer is only valid
```
Rustdoc will check the validity of the reference (i.e. ensure it isn't dangling) and then link to the item in the generated HTML documentation.

Make sure to apply the same to `finish()` below, and you may need to reflow the text to make the lines not be too long.

src/adapter/tracer/tests.rs
```diff
@@ -0,0 +1,226 @@
+// The Trustfall API requires the adapter to be passed in as an Arc.
+// Our adapter is not Send/Sync (it doesn't need it),
+// but there's currently nothing we can do about this lint.
+#![allow(clippy::arc_with_non_send_sync)]
```
obi1kenobi:
```suggestion
#![expect(clippy::arc_with_non_send_sync)]
```
so that clippy _asserts_ this lint is still here, or otherwise forces us to remove it

src/adapter/tracer/tests.rs
```diff
@@ -0,0 +1,226 @@
...
+    let tracing_adapter = Arc::new(TracingAdapter::new(&adapter));
+
+    let mut _results: Vec<_> =
+        trustfall::execute_query(&schema, tracing_adapter.clone(), query, variables.clone())
+            .expect("failed to run query")
+            .collect();
+
+    let tracer = tracing_adapter.finish();
+
+    let desired = vec![
```
obi1kenobi: Consider using a different test crate — possibly even a dedicated one for this test specifically. Our sealed trait logic is still imperfect and gets some edge cases wrong. It'd be really annoying if someone spends hours making trait analysis smarter, only to have to manually update this test which involves functionality they have likely never seen before, or at least not for months!

src/lib.rs
obi1kenobi: Of course. The best way to answer this is to look at how the user of the API will use it. I'd recommend cloning the `trustfall_rustdoc` repo, and using [cargo's `patch` functionality](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) to wire in the (not-yet-published) tracing code here into the actual use case. Even better: I'd recommend doing that all the way into `cargo-semver-checks` itself:
- have `cargo-semver-checks` patch `trustfall_rustdoc` to a local, unpublished copy
- have `trustfall_rustdoc` patch `trustfall-rustdoc-adapter` to a local, unpublished copy
- tweak the features `cargo-semver-checks` enables to only select one rustdoc version, for simplicity

For example, one issue I think that will quickly flag is that we currently use [one adapter for all parallel queries](https://github.com/obi1kenobi/cargo-semver-checks/blob/517ab562be8695d559e41edea13aa3d9a42b207b/src/check_release.rs#L349-L350), which might be a challenge given the use of `RefCell` here. The answer might be to clone the adapter inside `cargo-semver-checks`, or to be more clever here — I'm not sure what's easiest and the only way to find out is to try to make it work end-to-end.

In `cargo-semver-checks`, we'd eventually want something like a `--perf` flag which opts into collecting all these statistics, and then displays a useful data summary in the terminal (aggregated across all executed lints) while also saving any more detailed stats to file(s). It's your call how much of that vision you want to implement! But whatever we do here must be in line with making that vision possible, and ideally easy too :)

Let me know if you'd like to hop on a call to talk about this some more, or maybe even pair-program some of it.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,453 @@
...
+}
+
+impl Default for Tracer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[allow(clippy::enum_variant_names)] // Names match the adapter functions
```
obi1kenobi:
```suggestion
#[expect(clippy::enum_variant_names, reason = "names match the adapter functions")]
```

src/adapter/tracer/ptrace.rs
CLIDragon: Yes, I had it stored separately to avoid iterating when calculating `count`, but I now realise the overhead of incrementing a variable on every iteration will dominate that cost.

src/adapter/tracer/ptrace.rs
CLIDragon: To my mind, "how many vertices" reads as if we are counting the number of different vertices we resolved the property for, e.g. `Vid(1)` and `Vid(2)` and `Vid(3)`, not what we are counting which is number of calls to resolve a particular `(vid, type, property)` triple.

src/adapter/tracer/ptrace.rs
CLIDragon: For the first point, I was addressing that we could count how many `resolve_property` calls there were in total. Since we don't really care about that - it's an implementation detail of `trustfall`, we will not collect it. For the second, by 'results', I mean items in the iterator - we don't track how long is spent in calls to empty iterators.

I have updated it to make it clearer. In practice, I think since the comment applies broadly to all resolutions, it should be moved to a more global location, perhaps above the `TracingAdapter`.

src/lib.rs
CLIDragon: I'd love to get on a call to discuss, since the `c-s-c` end of the implementation has been a bit of a sticking point for me. I've sent you a message on Zulip with my availability - hopefully there's a time that works for both of us. 

I actually already have an end-to-end implementation working, though the implementation is a bit hackish. Because this is in `trustfall-rustdoc-adapter`, rather than `trustfall`, there's a bit of work that needs to be done in `trustfall_rustdoc` to get things working properly. Currently I have a mockup that looks like this:
```rs
pub fn trace_query<'slf: 'a, K: Into<Arc<str>>, V: Into<FieldValue>>(
    &'slf self,
    query: &str,
    vars: BTreeMap<K, V>,
) -> anyhow::Result<(Vec<QueryResult>, trustfall_rustdoc_adapter_v56::ptrace::Tracer)> {
    match self {
        #[cfg(feature = "v55")]
        VersionedRustdocAdapter::V55(_, adapter) => {
            execute_query(self.schema(), Arc::new(adapter), query, vars)
                .map(|x| x.collect())
                .map(|result_vec| (result_vec, trustfall_rustdoc_adapter_v56::ptrace::Tracer::new()))
        }

        #[cfg(feature = "v56")]
        VersionedRustdocAdapter::V56(_, adapter) => {
            let tracing_adapter = Arc::new(
                trustfall_rustdoc_adapter_v56::ptrace::TracingAdapter::new(adapter),
            );
            let ret = execute_query(self.schema(), tracing_adapter.clone(), query, vars)
                .map(|x| x.collect());

            let trace = tracing_adapter.finish();
            ret.map(|result_vec| (result_vec, trace))
        }
    }
}
```
which needs an additional versioned enum wrapper for the `Tracer`, since a concrete type won't work. I think cloning the `TracingAdapter` each time is a reasonable cost since we need a new `Tracer` anyway, so the only overhead is an extra `Arc`. Note that we can't directly use the `TracingAdapter` or `Tracer` inside of `cargo-semver-checks`, since we need a concrete type.

src/adapter/tracer/ptrace.rs
obi1kenobi: I do wish we had better terminology. But inside the scope of this method, we only have the ability to observe a single `Vid(_)` and the statement is unambiguous. We also have to assume that the person reading this code has some familiarity with the Trustfall adapter API already, which would help disambiguate further.

> How many times did we resolve the same property?

This on the other hand is actively inaccurate. If a user's query includes `@filter @output` or `@output @output(name: "foo")`, that user would be justifiably displeased when they see this count is _not actually_ "how many times did we resolve the same property" in their query.

In the Trustfall API, the "resolve" action happens _once total per API call_ not once per data vertex whose property is being loaded. Think of "resolve" as the thing that happens when you call a function in an interpreted language like Python: the runtime has to _resolve_ the name of the function, and then call it. If that function has a loop inside, we don't say we _resolved_ it more than once. Trustfall similarly makes _one_ call into the adapter. That adapter is our function; if it internally chooses to make many function calls under the hood, that is an implementation detail.

This documentation has to work for any implementation of the Trustfall adapter API, so:
- we only resolved the property once
- we count for how many vertices that happened

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,468 @@
...
+        type_name: &Arc<str>,
+        property_name: &Arc<str>,
+        resolve_info: &ResolveInfo,
+    ) -> ContextOutcomeIterator<'vertex, V, FieldValue> {
+        // TODO: Move this comment to a better location, to make it clear it applies
+        //      to all resolve_* functions (except resolve_starting_vertices).
+        // For each resolution we want to know:
+        // 1. What are we resolving?
+        // 2. How long did the resolution take?
+        // 3. How many times was a particular (vid, type, property) triple resolved?
```
obi1kenobi: See my other comment that used to be here. I don't think "resolve" is the right verb here.

Resolve = figure out whose responsibility it is to get the value we need. We only resolve once per Trustfall adapter API call.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,468 @@
...
+        //
+        // We are not collecting:
+        // (1) Number of times a function is called, and time spent constructing
+        //     the iterators.
+        //
+        //     The number of times any given resolution function is called is
+        //     an implementation detail of trustfall, and since most time is spent
+        //     resolving iterators, tracking it isn't valuable.
```
obi1kenobi: This feels pretty vague. "A function" presumably refers to Trustfall adapter API functions only, right?

It's also strange to have a numbered list, and then have item `(1)` have two things inside it.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// Create a new histogram.
+    pub fn new() -> ExpHistogram {
+        ExpHistogram { buckets: [0; 16] }
+    }
+
+    /// Add a value to the histogram.
+    pub fn add(&mut self, num: u64) {
+        for (i, lim) in BOUNDARIES.iter().enumerate() {
+            if num <= *lim {
+                self.buckets[i] = self.buckets[i].saturating_add(1);
```
obi1kenobi: `self.buckets[i]` has an implicit `panic!()` branch which will prevent Rust from vectorizing this loop.

Consider writing this with `self.buckets.iter_mut().zip(BOUNDARIES.iter())` instead to avoid the panic branch and (possibly) get vectorization.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// Returns the mean duration.
+    pub fn mean(&self) -> Duration {
+        // This cannot panic since count() is always at least 1.
+        self.sum / self.count()
+    }
+}
+
+/// Records and stores operations performed by the adapter.
+///
+/// This struct is intended for use inside of a TracingAdapter.
```
obi1kenobi:
```suggestion
/// This struct is intended for use inside of a [`TracingAdapter`].
```

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,468 @@
...
+            self.calls.insert(call_id.clone(), Summary::new(duration));
+        }
+    }
+
+    /// Set the duration of the last input.
+    pub fn record_last_input_duration(&mut self, duration: Duration) {
+        self.last_input_duration = Some(duration);
+    }
+
+    /// Get the duration of the last input. Panics if the duration is None.
```
obi1kenobi:
```suggestion
    /// Get the duration of the last input. Panics if the duration is `None`.
```

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,468 @@
...
+///
+/// This struct is intended for use inside of a TracingAdapter.
+/// Operations must be recorded sequentially in chronological order.
+/// Recording out-of-order operations will lead to invalid state.
+#[derive(Debug, Clone)]
+pub struct Tracer {
+    pub calls: BTreeMap<FunctionCall, Summary>,
+
+    // When we measure the time of an iterator, we also measure the time spent
+    // evaluating its inputs. We must therefore subtract this time out when
```
obi1kenobi: nit: possible confusion with "time out"
```suggestion
    // evaluating its inputs. We must therefore subtract out this time when
```

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    /// Returns the exponential histogram with times in nanoseconds.
+    pub fn histogram(&self) -> &ExpHistogram {
+        &self.hist
+    }
+
+    /// Returns the mean duration.
+    pub fn mean(&self) -> Duration {
+        // This cannot panic since count() is always at least 1.
```
obi1kenobi:
```suggestion
        // This cannot panic since `count()` is always at least 1.
```

src/adapter/tracer/ptrace.rs
CLIDragon: According to godbolt, https://godbolt.org/z/jsMKoMTbv, the compiler produces the exact same results in both cases.

obi1kenobi: That's fine---but note that neither Rust nor any other language guarantee that optimizations like this will continue to happen, or that they will always be applied on all platforms, or that they will continue to be applied as configuration flags change. So this statement

> the compiler produces the exact same results in both cases

is more accurately written

> the current version of the compiler, on `x86_64-unknown-linux-gnu`, with the specific configuration flags, produces the exact same results in both cases

`self.buckets.iter_mut().zip(BOUNDARIES.iter())` has no panic branches, on any platform under any configuration flags. It's a strictly lower-maintenance way to write the same code.

If you don't want to change it, I won't insist. But maintenance cost is important for an OSS project like this, and I'll probably get around to changing it myself eventually.

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,478 @@
...
+    FieldValue,
+    provider::{
+        Adapter, AsVertex, ContextIterator, ContextOutcomeIterator, EdgeParameters, Eid,
+        ResolveEdgeInfo, ResolveInfo, VertexInfo, VertexIterator, Vid,
+    },
+};
+
+/// A simple histogram that stores 15 roughly exponentially increasing buckets of
+/// values, from 0 to 1 billion, followed by a final bucket to store numbers greater
+/// 1 billion.
```
obi1kenobi:
```suggestion
/// than 1 billion.
```

src/adapter/tracer/ptrace.rs
```diff
@@ -0,0 +1,478 @@
...
+where
+    I: Iterator<Item = T>,
+    F: Fn(T, Duration) -> T,
+{
+    PerfSpanIter { inner, post_action }
+}
+
+/// Traces an inner adapter. Constructed with [TracingAdapter::new()]. Each
+/// tracer is only valid for one query. When switching between queries, if you
+/// use the same adapter, ensure you first call [TracingAdapter::finish()],
```
obi1kenobi:
```suggestion
/// Traces an inner adapter. Constructed with [`TracingAdapter::new()`]. Each
/// tracer is only valid for one query. When switching between queries, if you
/// use the same adapter, ensure you first call [`TracingAdapter::finish()`],
```

src/lib.rs
obi1kenobi: Wrapping the `Tracer` in another versioned wrapper seems reasonable. Would the same be required for the histograms and `Summary` etc.? That'd be a bit unfortunate, but I guess as a starting point it isn't terrible.

obi1kenobi (approved):

PR #1009, opened 2025-11-06
Add an index for `ImportablePath`

CLIDragon:
**Purpose of Change**
Tracing shows that `ResolveNeighbors(ImportablePath, Path)` can take a significant fraction of total runtime (5-10%).  

**Describe the Change**
Pre-calculate the importable path for every public item and replace path calculations with a lookup into a `HashMap`. 

**What alternatives have you considered?**
- The main speedup comes from eliminating the allocations of the paths, so if there was a way to do that without adding an index, it would be an easy win. 

**Additional Context**
This is a draft because I'm not sure of the performance impact. Last time I measured it, it came out neutral overall, but a new index and another dozen or so lints have been added since then.

### DIFF
--- a/src/adapter/edges.rs
+++ b/src/adapter/edges.rs
@@ -165,7 +165,7 @@ pub(super) fn resolve_importable_edge<'a, V: AsVertex<Vertex<'a>> + 'a>(
             match importable_paths {
                 Some(importable_paths) => Box::new(
                     importable_paths
-                        .into_iter()
+                        .iter()
                         .map(move |x| origin.make_importable_path_vertex(x)),
                 ),
                 None => Box::new(std::iter::empty()),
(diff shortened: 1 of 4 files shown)


obi1kenobi:
I'm a bit confused by this. Isn't the new index affecting `ResolveNeighbors(Importable, importable_path)` much more so than `ResolveProperty(ImportablePath, Path)`?

The [property resolution of `path` on `ImportablePath`](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/9c329559c9acb602192b9c43da4511d670b92730/src/adapter/properties.rs#L233-L243) remains unchanged here. It still does a bunch of `ToString::to_string` calls, allocates a `Vec` for them, and turns the whole thing into `Arc<[FieldValue]>` in the end.

CLIDragon:
> I'm a bit confused by this. Isn't the new index affecting `ResolveNeighbors(Importable, importable_path)` much more so than `ResolveProperty(ImportablePath, Path)`?

Good catch! I've updated the PR description to reflect the correct value.

Issue #687, opened 2024-12-23
Add `RawType` property to `FunctionParameter` type


suaviloquence:
I think in order to write correct witness functions in `c-s-c` we need to be able query the function signature more granularly than just the `signature` property.   Adding the ability to query the type of each parameter, in addition to the already-existing `GenericParameter` property, should be enough to do this (but I'm not 100% sure).  I can implement this, just opening an issue to make sure this is the right direction to go in.

obi1kenobi:
You seem to have a vision in mind here: could you share some more of it? I
haven't thought about the problem space nearly as much as you, and I'd love
to hear more of your thinking on it.

The idea seems reasonable in general, though I'm not super thrilled with
the current `RawType` vertex type design — I hacked it in as a workaround
originally and it hasn't seen any more love since then.

suaviloquence:
Current problem: we need to construct a witness function header (e.g., for function_missing, it needs to take parameters that have the same arguments/generics/generic bounds as the removed function). We can use `signature`, but it's kind of hacky, as the name of the removed function is hardcoded in the function signature, so we would have to print out a function like:

```rust
fn removed_function<'a, T>(x: &'a t, y: Cow<'static, str>) -> T::Item
where T: 'a + Iterator
{
    path::to::removed_function(x, y)
} 
```

which works, but may be confusing (the name of the witness function is the same as the removed function, and the return value is not technically necessary for a minimized withess?) -- another idea is to allow specifying a function name with `signature`, maybe?

Future problems: writing witnesses for functions that change arity/parameter types?

I was also under the assumption that `RawType` would eventually become (or be superseded by) a `Type` class that has the stringified `name` but also supports other queries (equality, satisfies trait bound, etc. - but this is for the future)

obi1kenobi:
All sounds good to me, go for it! Also, if you have opinions or ideas on how to design a better `Type` vertex type, I'd love to chat about that sometime :)

I'd be very curious to see what the witness query and template would look like for function arity change / parameter type change lints! I wonder if we might run into more issues there.

PR #1010, opened 2025-11-13
Add `type` property to `FunctionParameter` and `ReturnValue`

Hell-Tractor:
This pr is similar to #687.

At present, `FunctionParameter` only contains one property `name`. As I know, analyzing function signature is the only way to get parameter types and return value type.

Therefore these changes were made to support querying function parameter types and return value type directly. The new structure of `FunctionParameter` and `ReturnValue` is:

```graphql
type FunctionParameter {
    name: String!
    type: String!
}
type ReturnValue {
    is_unit: Boolean!
    type: String!
}
```

To make stringified `type` as similar to origin code as possible, it is converted from `rustdoc_types::Type` to `String` by using function `adapter::rust_type_name::rust_type_name`. The property `type` of function without return value is set to `()`.


test_crates/function_param_types_and_return_value_types/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+pub fn fn_with_generic_ref<T: std::ops::Add<Output = T>>(a: &T, b: &T) -> T
+where
+    T: Copy,
+{
+    *a + *b
+}
+
+pub fn fn_with_array_param(arr: [i32; 3]) -> i32 {
+    arr.iter().sum()
+}
```
obi1kenobi, line 35: Putting a single trailing newline at the end of files makes other tools work better.
```suggestion
}

```

If you use VSCode or a derived editor, I recommend adding the following to your config:
```
    "files.trimFinalNewlines": true,
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
```

src/adapter/tests.rs
```diff
@@ -8182,3 +8182,92 @@ fn function_return_value() {
 
     similar_asserts::assert_eq!(expected_results, results);
 }
+
+#[test]
+fn function_parameter_types_and_return_types() {
+    get_test_data!(data, function_param_types_and_return_value_types);
```
obi1kenobi, line 8188: This test is good. Would you mind adding another, separate test function that uses the `raw_type_json` test crate with this same query?

It has some much more complex types and we want to make sure every flavor of type we can think of is thoroughly covered by tests.

src/adapter/properties.rs
```diff
@@ -377,6 +385,15 @@ pub(super) fn resolve_return_value_property<'a, V: AsVertex<Vertex<'a>> + 'a>(
...
                 .into()
         }),
+        "type" => resolve_property_with(contexts, |vertex| {
+            vertex
+                .as_return_value()
+                .expect("not a return value")
+                .type_
+                .map(rust_type_name::rust_type_name)
+                .unwrap_or("()".to_string())
+                .into()
```
obi1kenobi: Let's try to avoid this `to_string()` call, because it'll be relatively common and we'll end up allocating twice (`.into()` will allocate again).

```suggestion
                .map(rust_type_name::rust_type_name)
                .map(Into::into)
                .unwrap_or("()".into())
```
Something like this should do the trick.

src/rustdoc_schema.graphql
```diff
@@ -1496,6 +1496,7 @@ https://docs.rs/rustdoc-types/0.11.0/rustdoc_types/struct.FnDecl.html
 """
 type FunctionParameter {
   name: String!
+  type: String!
```
obi1kenobi, line 1542: Is using `type` as an identifier syntactically valid here? Wow, I wouldn't have expected that. Nice!

Please make sure to add a thorough doc comment above it though? Unlike `name`, there are _many_ edge cases we need to write down here. Look at the `signature` doc comment as a starting point, though `type` certainly has more things to mention. For example:
- we do not specify whether types contained within use qualified paths or just names, nor a particular qualified path over another (e.g. `String` vs `std::string::String` vs `alloc::string::String`)
- generic parameters will appear without being defined and without their bounds, since the definition and bounds are syntactically part of the function, not the parameter

etc.

I recommend reading through the `raw_type_json` crate for some ideas how complex parameter types can get, and then writing down more docs :)

src/rustdoc_schema.graphql
```diff
@@ -2007,6 +2008,8 @@ type ReturnValue {
   ```
   """
   is_unit: Boolean!
+
+  type: String!
```
obi1kenobi, line 2098: We'll want a nice thorough doc comment here too.

obi1kenobi (commented):
Hey, thanks for the PR! Just a handful of suggestions about improving docs and testing and this should be good to merge.

If you don't mind me asking, I'm curious how you found this project and were inspired to open a PR! Are you already using it somewhere, or did it just seem interesting?

If you'd like to join the project's Discord, you're welcome to: https://discord.gg/wqhMJYfa

Hell-Tractor:
> If you don't mind me asking, I'm curious how you found this project and were inspired to open a PR! Are you already using it somewhere, or did it just seem interesting?

I'm a Master's students and my team work on detecting vulnerabilities on rust programs. My teammate found this project and this project is quite helpful for analyzing api docs of rust crates. We're trying using this project as part of our detection method. Our method will need type information of function input params and return value, so I tried to augment this project and this pr was made.

I have joined the Discord group and thank you for inviting me.

I'm now working on new  test cases and doc comment. Changes will be made later.

src/adapter/tests.rs
Hell-Tractor: Test function was add through commit f6e333c and 8c63736. The code seems working fine.

One thing to be noted:

In `raw_type_json/src/lib.rs`, parameter `d` in function `my_generic_function` is declared as:

```rust
d: impl for<'x> FnMut(
            &'a unsafe extern "C" fn(
                *const [u8],
                &'x mut *mut (),
                ...
            ) -> std::borrow::Cow<'static, [u8]>,
        ) -> &'x (dyn std::fmt::Display)
        + Send
        + 'static,
```

Stringified type of `d` is (I have added some white spaces manually to make it more readable):

```rust
impl for<'x> FnMut(
    &'a unsafe extern "C" fn(
        _: *const [u8],
        _: &'x mut *mut (),
        ...
    ) -> std::borrow::Cow<'static, [u8]>
) -> &'x (dyn std::fmt::Display)
+ Send
+ 'static
```

It's easy to notice that parameters of `unsafe extern "C" fn` was automatically added `_` as param names. Is this behavior expected?

src/rustdoc_schema.graphql
Hell-Tractor: I have added some doc comments through commit 957c0ae and 73e55be, specifying how type paths, generic parameters, life time parameters and `Self` type was stringified.

Is there anything more to be specified?

src/rustdoc_schema.graphql
Hell-Tractor: The doc comment here is almost same with the one in `FunctionParameter`. An additional note is included here stating that a unit return value will be stringified as `()`.

src/adapter/tests.rs
obi1kenobi: Is that syntax valid? I'd be surprised if so.

If it isn't valid syntax, we should probably fix it and make it emit correct syntax for function pointer types.

src/rustdoc_schema.graphql
```diff
@@ -1496,6 +1496,50 @@ https://docs.rs/rustdoc-types/0.11.0/rustdoc_types/struct.FnDecl.html
 """
 type FunctionParameter {
   name: String!
+
+  """
+  Stringified representation of the parameter's type, as close to the
+  original source code as possible.  May not preserve whitespace, constant
```
obi1kenobi, line 1502: nit: the docs style is to use single space between sentences
```suggestion
  original source code as possible. May not preserve whitespace, constant
```

src/rustdoc_schema.graphql
```diff
@@ -1496,6 +1496,50 @@ https://docs.rs/rustdoc-types/0.11.0/rustdoc_types/struct.FnDecl.html
...
+  expressions in types, or path names, but the output will be semantically
+  equivalent to the original.
+
+  The stringified type always within use full qualified path.
```
obi1kenobi, line 1506: "within" seems like a typo here.

Also, are you sure we can guarantee this in practice? If so, we'll need *a lot* more tests and documentation. Some of the edge cases we'll need to discuss:
- What if the type is defined at some path that is not publicly accessible, but it's re-exported at some other publicly importable path? Which fully-qualified path will be used?
- What if there's more than one fully-qualified path that is publicly importable, which one will be used? Concretely, why is `std::option::Option` chosen and not [`core::option::Option`](https://doc.rust-lang.org/core/option/enum.Option.html)?
- What if the type is defined in one crate and re-exported in another one, which fully-qualified path would we use? Does it matter if our crate has dependencies on both or only one of those crates where the type is defined or re-exported?

etc. etc.

This seems like a lot of work to define, test, and get merge-ready. In particular, we can't reasonably even say "we output whichever path rustdoc JSON gives us" because rustdoc JSON is an unstable output format — it definitely doesn't specify what path comes back, and it in principle could be different between different nightlies, on different platforms, or even between compilations on the same platform. So if we promise something stricter, we risk not being able to uphold it without a lot more work.

It may be better to offer less strict guarantees — but ones we can clearly define and which we know we can always uphold. For example, we could say it isn't well-defined whether fully-qualified or relative paths are used, or which path is used when there's more than one possibility. We can promise the shown path will be syntactically valid Rust, but not necessarily `pub`, not necessarily reachable from any particular location (not the same thing as `pub`), not necessarily shortest, not necessarily from the crate where the item was defined. It's _a_ way to refer to the item, but not any _specific_ way.

src/rustdoc_schema.graphql
```diff
@@ -1496,6 +1496,50 @@ https://docs.rs/rustdoc-types/0.11.0/rustdoc_types/struct.FnDecl.html
...
+  ```
+  fn generic_fn<T: Copy>(param: T) {}
+  ```
+
+  Lifetime parameters will be preserved.
+
+  For example, in the following code, the parameter type string is `&'a str`:
+  ```
+  fn lifetime_fn<'a>(param: &'a str) {}
+  ```
```
obi1kenobi, line 1531: How about in lifetime bounds like `+ '_`? These are common e.g. when returning iterators that borrow from `&self` but return copied/cloned values: the _iterator_ has lifetime `'_` ("the self lifetime") but the items it produces are owned.

src/rustdoc_schema.graphql
```diff
@@ -1496,6 +1496,50 @@ https://docs.rs/rustdoc-types/0.11.0/rustdoc_types/struct.FnDecl.html
...
+
+  `Self` types in methods will appear as `Self`.
+
+  For example, in the following code, the parameter type string is `Self`:
```
obi1kenobi, line 1535: It's a bit awkward to say "parameter type string" because there's risk of confusion with `String`, the type that a parameter can have.

How about we instead comprehensively rename `type` in the name of the property to `type_signature` (which is more clearly represented as a string) and then say "signature" here:
```suggestion
  For example, in the following code, the parameter type signature is `Self`:
```

src/rustdoc_schema.graphql
```diff
@@ -1496,6 +1496,50 @@ https://docs.rs/rustdoc-types/0.11.0/rustdoc_types/struct.FnDecl.html
...
+  ```
+  impl MyStruct {
+      fn method(self) {}
+  }
+  ```
+  """
+  type: String!
```
obi1kenobi, line 1542: Per my suggestion above, let's consider renaming this to `type_signature`?
```suggestion
  type_signature: String!
```

This would allow us to (in the future) add a `type` edge with more details (e.g. "is the type `Copy`" which has syntactic and SemVer implications).

Or if you feel like doing an A+ job in this PR, we could make that `type` edge immediately, and then add a new vertex type called `Type` with a property `signature` where all this would happen. Your call! If you don't feel like doing it right away, I'll probably refactor to this before too long.

test_crates/function_param_types_and_return_value_types/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+pub fn fn_with_qualified_param(p: std::option::Option<i32>) -> i32 {
+    p.unwrap_or(0)
+}
+
+use std::option::Option;
+pub fn fn_with_unqualified_param(p: Option<i32>) -> i32 {
+    p.unwrap_or(0)
+}
```
obi1kenobi, line 35: nit: we always leave a trailing line at the end of files, to improve interoperability with terminal tools (there's an IDE setting for this in all IDEs if you want to not check it manually)
```suggestion
}

```

src/rustdoc_schema.graphql
obi1kenobi: I unfortunately think there's more to be specified, and some things might be better if *explicitly* stated to not be guaranteed to happen one way or another. I left some more comments inline.

obi1kenobi (commented):
Thanks again for doing this work, and nice find on the bug with function pointer parameter types! That's definitely worth fixing, possibly in a separate PR if you feel like it.

Unfortunately there's more work on the doc comments and edge cases for writing out type signatures. I'm hoping it's one more iteration on the PR, though it might be more.

Also, this PR is pretty complex so I strongly recommend re-reviewing it for typos and flow after making a commit. It's easy for things to sneak by, and generally much easier to find those things when looking at the GitHub PR interface than one's own editor. Doing that will be a huge help for me, and will also help us both merge this faster.

obi1kenobi:
Hi! Just checking if you're still interested in driving this PR forward, or if you'd prefer if I get it over the finish line. I'm in no rush, just wanted to see if you still had the time and interest or if you got busy with other things :)

Hell-Tractor:
I'm busy on other works these days. I will be free in about a month. If you're in a hurry, then I'll have to trouble you to complete the rest of the parts.

obi1kenobi:
I'm happy to wait, just making sure you'll have the time and interest to
continue!

Enjoy the holidays :)

Hell-Tractor:
Same to you!

Merry Christmas XD

obi1kenobi:
Mostly superseded by #1064, so closing this. If you might still be interested in contributing, I'd love the help — you're welcome back anytime :)

PR #1076, opened 2026-06-15
Normalize associated item order to avoid wrong synthetic bound order.

obi1kenobi:
Ensure that normalization correctly handles associated items and will not accidentally rewrite bounds incorrectly. Fixes the bug identified in https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/1069#discussion_r3358407528

Also change how `impl Trait` in parameter position is represented, to avoid cases where adding an additional (nested) `impl Trait` in an earlier fn parameter changes the normalization numbering of `impl Trait` in later fn parameters, making it look like they changed types too. Now `impl Trait` in parameter position is represented with a parameter index and an occurrence index: `IT1_2` means this is the second parameter-position `impl Trait` of the first parameter in the function signature.

Exhaustively tested using `cargo-llvm-cov` and `cargo-mutants` in an AI-driven loop, adding more coverage and fixing bugs until fixpoint.

### DIFF
--- a/src/adapter/normalize/context.rs
+++ b/src/adapter/normalize/context.rs
@@ -1,13 +1,11 @@
-use std::{borrow::Cow, num::NonZeroUsize};
-
-use rustdoc_types::{
-    AssocItemConstraint, AssocItemConstraintKind, FunctionSignature, GenericArg, GenericArgs,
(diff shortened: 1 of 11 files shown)


obi1kenobi:
@codex review please

PR #1080, opened 2026-06-30
Allow std library stability to influence what is public API.

obi1kenobi:
In preparation for allowing `cargo-semver-checks` to check the Rust standard library for accidental breakage, add a mode where stability affects what is considered public API. In that mode:
- unstable items are not public API
- stable items that are const-unstable have their const-ness hidden, as it isn't public API

As stability attributes are not stable, the regular mode of the adapter ignores stability for public API purposes.

Additionally, we fix a bug in `#[doc(hidden)]` propagation with respect to `#[doc(hidden)]` re-exports: items made available as a result of a `#[doc(hidden)]` glob re-export are not public API even though the final path might have no `#[doc(hidden)]` components. In other words, causing a path to exist as a result of `#[doc(hidden)]` always renders the path non-public API.

An example of this is:
```rust
mod example {
    #[doc(hidden)]
    pub use inner::*;

    pub mod inner {
        pub struct Example;
    }
}

pub use example::*;
```

Here, `this_crate::Example` is re-exported at top level, and the components of the path _appear to be_ public API: `this_crate` is a crate and therefore implicitly `pub`, while `Example` is explicitly `pub`, and neither is `#[doc(hidden)]`. However, the hidden glob re-export is what makes `this_crate::Example` available, so that name is not public API.

This bug was spotted by @codex.

### DIFF
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -117,7 +117,7 @@ impl<'a> Adapter<'a> for &'a RustdocAdapter<'a> {
             match type_name.as_ref() {
                 "Crate" => properties::resolve_crate_property(contexts, property_name),
                 "Item" | "Importable" | "GenericItem" => {
-                    properties::resolve_item_property(contexts, property_name)
+                    properties::resolve_item_property(contexts, property_name, self)
                 }
                 "ImplOwner"
                 | "Struct"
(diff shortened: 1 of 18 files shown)


obi1kenobi:
@codex review please

PR #1087, opened 2026-07-02
Support default-value stability in rustdoc JSON v60.

obi1kenobi:
Unstable default values, such as provided function bodies in trait definitions, associated type defaults, and default values for associated constants, are treated as not present when indexing rustdoc JSON in "standard library mode".

This additionally has implications on trait sealing: a stable implementation of a trait may be forced to provide a value or implementation for associated items, which can cause some traits to be considered public API sealed within the stable public API. See the test cases for examples of this.


obi1kenobi:
@codex review please

PR #1088, opened 2026-07-02
Normalize parent, item, and higher-ranked generics separately.

obi1kenobi:
Previously, adding a generic param to a parent `impl` scope could cause the normalized generic params of functions in that `impl` to get renumbered, making it look like those functions' signatures had themselves changed as well.

This PR separates the normalization for parent, item, and higher-ranked generics so that they all use separate namespaces, and can grow and shrink independently. This will cut down on unnecessary (and expensive!) type checks and will make `cargo-semver-checks` lints more efficient.

Supersedes #1086 which was irreversibly closed due to what appears to be a GitHub bug.


obi1kenobi:
@codex review please

obi1kenobi:
@codex review please

PR #1107, opened 2026-07-20
Add @transform to graphql schema

ivelieu:
As discussed on the discord, this is a one line change I made to fix erroneous LSP errors over trustfall queries.

### DIFF
--- a/src/rustdoc_schema.graphql
+++ b/src/rustdoc_schema.graphql
@@ -1,9 +1,9 @@
 schema {
   query: RootSchemaQuery
 }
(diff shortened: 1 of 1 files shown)


src/rustdoc_schema.graphql
obi1kenobi: These three should also be `repeatable` btw. If this parses with the current parser library (it didn't use to on a prior version), we should be good to add all of them.

obi1kenobi (approved):

obi1kenobi:
I need to fix nightly first it seems.

obi1kenobi:
Thanks for the PR and for the patience while I unbroke nightly and merged it!

PR #1113, opened 2026-08-02
Add `generic_parameter` node to `AssociatedType`

ivelieu:
For use with https://github.com/obi1kenobi/cargo-semver-checks/issues/1660 to lint against generic associated types on traits.

I've added a test for when the GAT contains lifetimes, types or a paramaterized const.

### DIFF
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -337,7 +337,7 @@ impl<'a> Adapter<'a> for &'a RustdocAdapter<'a> {
                 edges::resolve_normalized_type_signature_edge(contexts, edge_name)
             }
             "GenericItem" | "ImplOwner" | "Struct" | "Enum" | "Union" | "Trait" | "Function"
-            | "Method" | "Impl"
+            | "Method" | "Impl" | "AssociatedType"
                 if matches!(edge_name.as_ref(), "generic_parameter") =>
             {
                 edges::resolve_generic_parameter_edge(contexts, edge_name)
@@ -405,7 +405,13 @@ impl<'a> Adapter<'a> for &'a RustdocAdapter<'a> {
                         ),
                         "GenericItem" => matches!(
                             actual_type_name,
-                            "Struct" | "Enum" | "Union" | "Trait" | "Function" | "Method"
+                            "Struct"
+                                | "Enum"
+                                | "Union"
+                                | "Trait"
+                                | "Function"
+                                | "Method"
+                                | "AssociatedType"
                         ),
                         "Variant" => matches!(
                             actual_type_name,
--- a/src/adapter/tests.rs
+++ b/src/adapter/tests.rs
@@ -13236,3 +13236,156 @@ fn function_parameter_normalized_type_signature_lint_shape() {
 
     similar_asserts::assert_eq!(expected_results, results);
 }
+
+#[test]
+fn rustdoc_trait_has_generic_associated_types() {
+    get_test_data!(data, generic_associated_types);
+    let adapter = RustdocAdapter::new(&data, None);
+    let adapter = Arc::new(&adapter);
+
+    let query = r#"
+{
+    Crate {
+        item {
+            ... on Trait {
+                trait_name: name @output
+                associated_type {
+                    type_name: name @output
+                    generic_parameter {
+                        generic_name: name @output
+                    }
+                }
+            }
+        }
+    }
+}
+"#;
+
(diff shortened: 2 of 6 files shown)


src/rustdoc_schema.graphql
```diff
@@ -2607,6 +2607,12 @@ type AssociatedType implements Item {
   """
   has_default: Boolean!
 
+  # edges from GenericItem
+  """
+  Generic type, lifetime, or const parameters by which this associated trait is parameterized.
```
```suggestion
  Generic type, lifetime, or const parameters by which this associated type is parameterized.
```

test_crates/generic_associated_types/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "generic_associated_types"
+version = "0.1.0"
+edition = "2024"
```
obi1kenobi, line 4:
```suggestion
edition = "2024"
publish = false
```

src/adapter/mod.rs
```diff
@@ -337,7 +337,7 @@ impl<'a> Adapter<'a> for &'a RustdocAdapter<'a> {
                 edges::resolve_normalized_type_signature_edge(contexts, edge_name)
             }
             "GenericItem" | "ImplOwner" | "Struct" | "Enum" | "Union" | "Trait" | "Function"
-            | "Method" | "Impl"
+            | "Method" | "Impl" | "AssociatedType"
```
obi1kenobi, line 340: I think you'll also need to update the type coercion code to match.

Otherwise querying `Item -> ... on GenericItem -> generic_parameter` won't return any GATs or generic params.

Please make sure to add a regression test here.

test_crates/generic_associated_types/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+use std::{ops::Deref, rc::Rc};
```
obi1kenobi: I think `std::rc::Rc` is an unused import. Please make sure the test crates aren't linty.

obi1kenobi (commented):
Broadly directionally good, but there are a handful of issues to fix before we merge.

Please be particularly careful with copy-paste errors. Consider self-reviewing your PR's code using the GitHub UI to find such issues, because they are easier to see with the "change of scenery" from your editor.

test_crates/generic_associated_types/Cargo.toml
obi1kenobi: This one is on me for not spelling it out in the contributing docs, which I've now fixed. Basically we don't want to accidentally publish these crates, so we mark them unpublishable.

test_crates/generic_associated_types/src/lib.rs
ivelieu: sorry! one idea I'm thinking of trying to avoid this in future is to have an external LSP-checked crate symlink into the repo's test crate file, and editing through that so I get the LSP lints without copy/pasting to avoid doing this accidentally in the future.

test_crates/generic_associated_types/Cargo.toml
ivelieu: thank you, I didn't know that was an option.

src/rustdoc_schema.graphql
ivelieu: apologies, that was just my confusing the GAT acronym ><

src/adapter/mod.rs
ivelieu: done, test added.

src/adapter/tests.rs
```diff
@@ -13236,3 +13236,156 @@ fn function_parameter_normalized_type_signature_lint_shape() {
...
+fn rustdoc_item_has_generic_associated_types() {
+    get_test_data!(data, generic_associated_types);
+    let adapter = RustdocAdapter::new(&data, None);
+    let adapter = Arc::new(&adapter);
+
+    let query = r#"
+{
+    Crate {
+        item {
+        ... on GenericItem {
```
obi1kenobi:
```suggestion
            ... on GenericItem {
```

obi1kenobi (approved):
Nice, approving and merging after a small indentation fix. I'll also cut a new release of this library so you can use this in cargo-semver-checks.

Projects I manage directly generally squash-merge, so they don't require anything special in terms of git history. Force-pushes generally make code review harder because the GitHub UI makes it hard to double-check that past feedback has been acted on (since the prior commit is gone and there's no diff from it), so I usually recommend not using force-pushes if they can be avoided.

